PR #1, opened 2019-10-24
Development and production build improvements

santhoshtr:
* Introduce hot reload when `npm run dev` used. Will be helpful
  while developing.
* Use CSS loader with webpack instead of compiling with less
* Include style in the index.js, so that dependencies are tracked
  by webpack loaders
* Minor: package.json private value from string to boolean
* Move all style in to source folder.


stephanebisson:
`npm run dev` is not working for me. I don't know what I am missing. It is always serving the content of the `dist/` folder (as specified in index.html). Is the dev-server serving the live-compiled assets from different URLs, should the index.html be different?

santhoshtr:
what happens when you run `npm run dev`. Not that there a few added depdencies, so you will need to install them first.
This is the expected result after running `npm run dev`
image

stephanebisson:
Yep, I get the same thing in the shell but when I go to that URL, I get the index.html that is pulling the assets from the `dist/` folder generated from the build command. 

Try modifying any style and you'll see that it is being picked up by webpack-dev-server but you don't see it in the app. Another way to test this is to delete the dist folder first and then trying to run the app with the dev command.

santhoshtr:
I realized that this application is not bootstrapped using preact-cli. That is why many of these connections are missing and we need to manually set up.

Consider using preact-cli to bootstrap an application and then add the code so far developed to it. That saves lot of time setting up this developer tooling and so on.

PR #2, opened 2019-11-07
Cypress tests

jpita:
adding first cypress tests


cypress/integration/spec.js
```diff
@@ -0,0 +1,34 @@
...
+  })
+
+  it('search should show results', () => {
+    cy.get('.result').first()
+      .should('have.text', "Cattaraugus County, New YorkCounty in New York")
+  })
+
+  it('results with image should show image', () => {
+    cy.get('.result').first()
+      .children().should('have.length', 2)
```
stephanebisson: The exact number of children is something that can change for unrelated reasons. For instance: it is common to change the HTML structure to accomplish different alignment requirements. 

A more stable way to test could be to check the presence of an  tag inside the search result row.

cypress/integration/spec.js
```diff
@@ -0,0 +1,34 @@
...
+  })
+
+  it('results without image should not show image', () => {
+    cy.get('.result').first().next()
+      .children().should('have.length', 1)
```
stephanebisson: Previous comment also applies here.

cypress/integration/spec.js
```diff
@@ -0,0 +1,34 @@
...
+  })
+  
+  it('article should open from search results page', () => {
+    cy.get('.result').first().next()
+    .children().should('have.length', 1)
+    cy.get('body').type('{enter}').type('{downarrow}').type('{enter}')
+    cy.get('.title').should('have.text', "Cattaraugus County, New York")
+  })
+})
```
stephanebisson: All files should end with an empty line.

cypress/integration/spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+  it('results without image should not show image', () => {
+    cy.get('.result').first().next()
+      .children().should('have.length', 1)
+  })
+  
+  it('article should open from search results page', () => {
+    cy.get('.result').first().next()
+    .children().should('have.length', 1)
+    cy.get('body').type('{enter}').type('{downarrow}').type('{enter}')
```
stephanebisson: Those keys are going to be used for all interactions. It might be worth encapsulating them in a specific class similar to a page object. Doesn't have to be in the PR, just something to consider instead of retyping the same kind of line everywhere.

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
+// This example commands.js shows you how to
+// create various custom commands and overwrite
+// existing commands.
+//
+// For more comprehensive examples of custom
+// commands please read more here:
+// https://on.cypress.io/custom-commands
```
stephanebisson: Custom commands could be an interesting way to encapsulate the device keypad. We could then use them like so:
cy.left();
cy.enter();

README.md
```diff
@@ -3,3 +3,9 @@
 The official Wikipedia KaiOS app.
 
 Tasks and issues are tracked on [Phabricator](https://phabricator.wikimedia.org/project/profile/4305/).
+
+## To run the Cypress tests
+### To run headlessly
+`npm test`
```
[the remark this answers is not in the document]
jpita: maybe we can use a combination of [these](https://askubuntu.com/a/539293)?
`npm run build && npm run start & npm test`

cypress/integration/spec.js
```diff
@@ -0,0 +1,34 @@
...
+// @ts-check
+const searchPage = new SearchPage()
+describe('Article search', () =>{
+  beforeEach(() =>{
+    searchPage.navigateToSearchPage()
+    searchPage.search("catt")
+  })
+
+  it('search should show results', () => {
+    cy.get('.result').first()
```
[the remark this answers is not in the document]
jpita: fixed

cypress/integration/spec.js
jpita: fixed

cypress/integration/spec.js
jpita: fixed

cypress/integration/spec.js
```diff
@@ -0,0 +1,34 @@
...
+    searchPage.search("catt")
+  })
+
+  it('search should show results', () => {
+    searchPage.results().first()
+      .should('have.text', "Cattaraugus County, New YorkCounty in New York")
+  })
+
+  it('results with image should show image', () => {
+    cy.get('.result').first()
```
stephanebisson: All instances of `cy.get('.result')` should also be updated to use `searchPage.results()`

cypress/integration/spec.js
```diff
@@ -0,0 +1,34 @@
...
+  it('results without image should not show image', () => {
+    cy.get('.result').first().next()
+      .children().should('have.length', 1)
+  })
+  
+  it('article should open from search results page', () => {
+    cy.get('.result').first().next()
+    .children().should('have.length', 1)
+    cy.get('body').type('{enter}').type('{downarrow}').type('{enter}')
+    cy.get('.title').should('have.text', "Cattaraugus County, New York")
```
stephanebisson: This looks like it's looking at the article page. Do you want a new page object for that?

cypress/integration/spec.js
jpita: was fixed before but forgot to push the commit

PR #7, opened 2019-11-19
Add webpack-dev-server to server localhost, and listen the changes

hueitan:
all the dev server configuration can be done in webpack config file, listen to the file changes but doesn't support hot reloading (not sure whether needed, will revisit if needed)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -993,6 +993,35 @@
         }
       }
     },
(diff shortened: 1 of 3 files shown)


hueitan:
Fixed @stephanebisson, thanks for spotting the error of the configuration.

PR #8, opened 2019-11-21
Making <SoftKey /> component always stay on the screen

hueitan:
It covers all the SoftKey changes in current status, let me know if anything else breaks.
Do also check the device performance.


stephanebisson:
In a follow up we could consider exposing an interface that is higher level than `dispatch` so that components can update the softkeys in one call.

PR #9, opened 2019-11-22
Abstract <ListView/> Component

hueitan:
> props : 
> { items : { title, titleHtml, description, imageUrl }, containerRef }

### DIFF
--- a/cypress/page-objects/search-page.js
+++ b/cypress/page-objects/search-page.js
@@ -8,6 +8,6 @@ export class SearchPage {
     }
 
     results(){
-        return cy.get('.result')
+        return cy.get('.list')
     }
 }
--- a/cypress/page-objects/search-page.js
+++ b/cypress/page-objects/search-page.js
@@ -8,6 +8,6 @@ export class SearchPage {
     }
 
     results(){
-        return cy.get('.list')
+        return cy.get('.item')
     }
 }
--- a/src/components/ListView.js
+++ b/src/components/ListView.js
@@ -6,10 +6,10 @@ import { h } from 'preact'
  */
 export const ListView = ({ items, containerRef }) => {
   if (!items.length) return
-  return <div class='results' ref={containerRef}>
+  return <div class='list' ref={containerRef}>
     {
       items.map(item => (
-        <div class='result' data-selectable data-title={item.title} data-selected-key={item.title} key={item.title}>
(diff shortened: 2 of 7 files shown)


src/components/ListView.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * The containerRef is suggested to be used together with the hooks useNavigation
+ * without the containerRef, the view won't scroll to the selected row
+ */
+export const ListView = ({ items, containerRef }) => {
+  if (!items.length) return
+  return <div class='results' ref={containerRef}>
+    {
+      items.map(item => (
+        <div class='result' data-selectable data-title={item.title} data-selected-key={item.title} key={item.title}>
```
stephanebisson: The assumed structure of item correspond to a search results item. Do you plan to factor out the details of `item` out of this component later or do you anticipate that this structure will actually work for all ListView usages?

style/listview.less
```diff
@@ -0,0 +1,59 @@
...
+
+.result .img img {
+	width: 38px;
+	height: 38px;
+	border-radius: 4px;
+}
+
+.result[nav-selected="true"] {
+	background-color: #eaf3ff;
+}
```
stephanebisson: Missing a line return here. 

I wish linting could catch that.

src/components/ListView.js
hueitan: currently, what I foresee the common list view in KaiOS, items are all selectable, with the title, optional description, optional image.

I think we can do more changes when we use it on the other pages with different requirements.

style/listview.less
hueitan: not sure how many linting tools for less, maybe can try this https://github.com/lesshint/lesshint

out of scope now, can try doing this in other pr.

cypress/page-objects/search-page.js
```diff
@@ -8,6 +8,6 @@ export class SearchPage {
     }
 
     results(){
-        return cy.get('.result')
+        return cy.get('.list')
```
stephanebisson: This should probably be `.item`

stephanebisson:
Something to think about for the future: can we encapsulate scrolling and positioning inside the ListView component.

hueitan:
Yes, true, was thinking the same, the scrolling behavior should be handled in the listview component itself. will look into it.

PR #11, opened 2019-11-25
Add missing translation key in qqq.json following the previous commit

hueitan:


i18n/qqq.json
```diff
@@ -9,6 +9,7 @@
 	"app-subtitle": "Subtitle of the application. Shown in the app store.",
 	"app-description": "Description of the application. Shown in the app store.",
 	"search-placeholder": "Placeholder of the search input.\n{{Identical|Search}}",
+	"settings": "Settings",
```
stephanebisson: This should be a sentence describing where it is being used. When it's displayed on a button (or mapped to a key, like here) it is also useful to say what the button does so that translators can try to find the appropriate term.

PR #12, opened 2019-11-25
Add <Settings /> page

hueitan:
<s>🚧 Don't merge this first, this wip is used for the reviewing and testing purpose </s>

There are several things to be done

- [x] Translation for the title
- [x] Page header title "Settings"
- [x] <s>scroll to view "useNavigation"</s> _Invalid_ 
- [x] <s>Enter to link new page</s> _out of scope now_
- [x] TBC

The page route to Settings page instead opening a pop up, can change it in the future.


hueitan:
The update of the latest commit only customize the Settings page styling (same styling as section page), the search page has different styling, I think we can let it go now until we have more pages then making the styling more generic.

PR #14, opened 2019-11-27
Go to article from search page

apap04:
This allows the user to go to the article from searching. Unfortunately, any other page that uses `ListView` will also have this article linking.

### DIFF
--- a/src/components/ListView.js
+++ b/src/components/ListView.js
@@ -10,10 +10,10 @@ export const ListView = ({ items, containerRef }) => {
     {
       items.map(item => (
         <div class='item' data-selectable data-title={item.title} data-selected-key={item.title} key={item.title}>
-          <div class='info'>
+          <a class='info' href={`/article/en/${item.title}`}>
             <div class='title' dangerouslySetInnerHTML={{ __html: item.titleHtml || item.title }} />
             { item.description && <div class='description'>{item.description}</div> }
-          </div>
+          </a>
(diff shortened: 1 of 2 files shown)


src/components/ListView.js
```diff
@@ -10,10 +10,10 @@ export const ListView = ({ items, containerRef }) => {
     {
       items.map(item => (
         <div class='item' data-selectable data-title={item.title} data-selected-key={item.title} key={item.title}>
-          <div class='info'>
+          <a class='info' href={`/article/en/${item.title}`}>
```
hueitan: 👍 I would love to see more opportunities for using html markup over programmatic.

I have tried removing all the relevant event here, it's only working when you `Tab and Enter` or `click' on the desktop browser, doesn't work in KaiOS.

does it work for you?

apap04: > 👍 I would love to see more opportunities for using html markup over programmatic.
> 
> I have tried removing all the relevant event here, it only works when you `Tab and Enter` or `click' on the desktop browser, doesn't work in KaiOS.
> 
> does it work for you?

I'll get my emulator working on my Linux machine soon to test the changes further..

apap04: bah, this change was intended for links to work on the Search page.. it's already implemented [here](https://github.com/wikimedia/wikipedia-kaios/blob/f831a31b165b4ba2288d1e11a93adb0b1cea05b0/src/components/Search.js#L32).

PR #15, opened 2019-11-29
Render the content section by section

hueitan:
**Before :** Render lead section and all remaining section
**After :** Render lead as the first page, then render the next section with toclevel 1 when continue scrolling

This change updates the **render section by section** when user scroll to the last page.

When a user scrolls up when at the top of the section, will render the previous section's first page, the behavior is not very clear now, therefore we use this solution until we know it.

### DIFF
--- a/src/api/wiki.js
+++ b/src/api/wiki.js
@@ -48,7 +48,9 @@ const getArticle = (lang, title) => {
         nextContent = ''
       }
 
-      if (s.toclevel === 1) nextDescription = s.line
+      if (s.toclevel === 1) {
+        nextDescription = s.line
+      }
 
       const header = 'h' + (s.toclevel + 1)
       const headerLine = `<${header}>${s.line}</${header}>`
--- a/src/api/wiki.js
+++ b/src/api/wiki.js
@@ -30,6 +30,7 @@ const getArticle = (lang, title) => {
     // parse lead as the first section
     const sections = []
     sections.push({
+      imageUrl,
       description: data.lead.description,
       content: data.lead.sections[0].text
     })
@@ -40,8 +41,14 @@ const getArticle = (lang, title) => {
     data.remaining.sections.forEach((s) => {
       // the section starts with every toclevel 1
       if (s.toclevel === 1 && nextDescription) {
+        // search for the first image in the content
+        const doc = parser.parseFromString(nextContent, 'text/html')
+        const imgNode = doc.querySelector('img')
+
         sections.push({
-          description: nextDescription, content: nextContent
+          description: nextDescription,
+          content: nextContent,
+          imageUrl: (imgNode && imgNode.getAttribute('src')) || imageUrl
         })
 
         nextDescription = ''
@@ -65,7 +72,6 @@ const getArticle = (lang, title) => {
 
     return {
       title,
-      imageUrl,
       sections,
       infobox
     }
--- a/src/api/wiki.js
+++ b/src/api/wiki.js
@@ -19,7 +19,6 @@ const getArticle = (lang, title) => {
   const url = `https://${lang}.wikipedia.org/api/rest_v1/page/mobile-sections/${title}`
   return cachedFetch(url, data => {
     const parser = new DOMParser()
-    const title = data.lead.displaytitle
     const imageUrl = data.lead.image && data.lead.image.urls['320']
 
     // parse info box
@@ -31,32 +30,33 @@ const getArticle = (lang, title) => {
     const sections = []
     sections.push({
       imageUrl,
+      title: data.lead.displaytitle,
       description: data.lead.description,
       content: data.lead.sections[0].text
     })
 
     // parse section as the remaining section
-    let nextDescription = ''
+    let nextTitle = ''
     let nextContent = ''
     data.remaining.sections.forEach((s) => {
       // the section starts with every toclevel 1
-      if (s.toclevel === 1 && nextDescription) {
+      if (s.toclevel === 1 && nextTitle) {
         // search for the first image in the content
         const doc = parser.parseFromString(nextContent, 'text/html')
         const imgNode = doc.querySelector('img')
 
         sections.push({
-          description: nextDescription,
+          title: nextTitle,
           content: nextContent,
           imageUrl: (imgNode && imgNode.getAttribute('src')) || imageUrl
         })
 
-        nextDescription = ''
+        nextTitle = ''
         nextContent = ''
       }
 
       if (s.toclevel === 1) {
-        nextDescription = s.line
+        nextTitle = s.line
       }
 
       const header = 'h' + (s.toclevel + 1)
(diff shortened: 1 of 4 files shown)


src/hooks/useScroll.js
```diff
@@ -3,15 +3,32 @@ import { useKeys } from 'hooks'
...
     },
     ArrowUp: () => {
+      const previous = elementRef.current[prop]
       elementRef.current[prop] -= step
+      const after = elementRef.current[prop]
+
+      // show the previous page (section in article)
+      if (previous === after) {
+        event.prevPage()
+        elementRef.current[prop] = 0
```
hueitan: _here we show the location 0, as we don't have previous section information (can only know this after the section is being rendered)._

src/components/Article.js
```diff
@@ -43,22 +48,32 @@ export const Article = ({ lang, title }) => {
...
   return (
     <Fragment>
-      <Pager>
+      <Pager event={pagerEvent}>
         <div class='page article'>
           { hasImage && <div class='lead-image' style={{ backgroundImage: `url(${article.imageUrl})` }} /> }
           <div class={'card' + (hasImage ? ' with-image' : '')}>
             <div class='title' dangerouslySetInnerHTML={{ __html: article.title }} />
-            { article.description && (
+            { article.sections[currentSection].description && (
```
hueitan: _the `title` always keep the same, the `description` shows the `lead description` or the section `header text`_

the actual design of this will be defined later.

src/components/Article.js
```diff
@@ -43,22 +48,32 @@ export const Article = ({ lang, title }) => {
...
-  }
-
   const hasImage = !!article.imageUrl
 
+  const pagerEvent = {
+    nextPage: () => {
+      const sectionLength = article.sections.length
+      const nextSection = currentSection + 1
+
+      if (nextSection < sectionLength) setCurrentSection(nextSection)
```
stephanebisson: Please always use curly braces with conditions.

In this case, however, it can be simplified to
`setCurrentSection( nextSection < sectionLength ? nextSection : 0 )`

src/hooks/useScroll.js
stephanebisson: We'll have to solve that. It's really confusing.

src/components/Article.js
```diff
@@ -43,22 +48,32 @@ export const Article = ({ lang, title }) => {
...
+  }
+
   return (
     <Fragment>
-      <Pager>
+      <Pager event={pagerEvent}>
         <div class='page article'>
           { hasImage && <div class='lead-image' style={{ backgroundImage: `url(${article.imageUrl})` }} /> }
           <div class={'card' + (hasImage ? ' with-image' : '')}>
             <div class='title' dangerouslySetInnerHTML={{ __html: article.title }} />
```
stephanebisson: The title should be the title of the article on the first page but then the title of the section.

src/components/Article.js
```diff
@@ -43,22 +48,32 @@ export const Article = ({ lang, title }) => {
...
+      if (prevSection >= 0) setCurrentSection(prevSection)
+    }
+  }
+
   return (
     <Fragment>
-      <Pager>
+      <Pager event={pagerEvent}>
         <div class='page article'>
           { hasImage && <div class='lead-image' style={{ backgroundImage: `url(${article.imageUrl})` }} /> }
```
stephanebisson: The lead image should change to be the first image found in each section.

src/api/wiki.js
```diff
@@ -19,30 +19,51 @@ const getArticle = (lang, title) => {
...
-      content += headerLine
-      content += s.text
-    })
 
-    // The app is served from the app:// protocol so protocol-relative
-    // image sources don't work.
-    content = content.replace(/src="\/\//gi, 'src="https://')
+      // The app is served from the app:// protocol so protocol-relative
+      // image sources don't work.
+      nextContent += headerLine + s.text.replace(/src="\/\//gi, 'src="https://')
```
stephanebisson: The header should not be part of the content since it will be show as the title on the title page.

src/components/Article.js
hueitan: what should we show the description in the remaining section?

src/api/wiki.js
hueitan: currently the section starts with `toclevel 1`, `toclevel 2` and the following will grouped in the previous `toclevel 1`, 

I will change the current behaviour to include the headerLine only when the toclevel is not 1, let me know if this is fine.

hueitan:
> In general, the relationship between Article, Pager, and useScroll could be improved. It's becoming confusing. We can think about that and refactor as a second step.

Agree, that is too confused when developing it, here's my thought

> useScroll
> --- useful when the page is able to scroll to a given unit when scroll up down / left right
> --- use case : multiple page, the view cannot fit into one page

> useNavigation
> --- useful when there are selection elements, and navigate to view
> --- use case : list view, a list of choices

Usually, we will need to define either one or both for the pages when the view cannot fit into one page. Therefore, `useScroll` and `useNavigation` are recommended to be defined in every view (Article, Search, Settings, etc).

For this specific `Pager`, we don't need `Pager` here, we define the `useScroll` in article page, then things got clearer

src/components/Article.js
hueitan: see https://github.com/wikimedia/wikipedia-kaios/pull/15/commits/c823f06edc11a023b89ccda5cf7723fc06bad250

src/components/Article.js
hueitan: done in https://github.com/wikimedia/wikipedia-kaios/pull/15/commits/c15095c1867629dc484e528377e0187ef2c37a14

src/components/Article.js
stephanebisson: Please show the section title as the big title and leave the description and line out of the section title pages.

hueitan: https://github.com/wikimedia/wikipedia-kaios/pull/15/commits/b6f4961ca47fc76648a4e9f449224b4552d6806c

Lead view
Screen Shot 2019-12-03 at 1 21 57 PM

Section view
Screen Shot 2019-12-03 at 1 22 05 PM

src/hooks/useScroll.js
hueitan: There is a quick and dirty solution for this

```js
setTimeout(() => {
  const scrollWidth = elementRef.current.scrollWidth
  const offset = 240
  elementRef.current[prop] = scrollWidth - offset
}, 0)
```

The real challenging we have here is that we render section by section, we don't have the previous section in the view, therefore no scrollWidth information. The current workflow is 

1. Render the section 
2. set the current scrollLeft

Therefore, this quick solution, you will see the first page of the section, then immediately to the last page of that section. 

Or we might need to use an extra layer (loading) when doing the transition.

stephanebisson: If we do the second step inside `useLayoutEffect` instead of `useEffect` there will be no flash between the steps.

hueitan: Thanks!! Let me try that.

PR #17, opened 2019-12-02
test for back button action

jpita:


cypress/integration/spec.js
```diff
@@ -36,4 +36,16 @@ describe('Article search', () =>{
...
+  it('back button should take us to search page with results', () => {
+    searchPage.search("cattle")
+    searchPage.results().first()
+    cy.enter()
+    cy.downArrow()
+    cy.enter()
+    articlePage.title().should('have.text', "Cattle")
+    cy.clickCloseButton()
+    searchPage.results().first()
+      .children().first().next().should('have.class', 'img')
```
hueitan: how do we tell having the class img belongs to search page with results?

jpita: I'm sorry, I don't understand the question

jpita:
> The invisionapp sketch shows the `left / center / right` are `menu / selection / sections`
> 
> I think there is no close button in this view, therefore the default back button is needed.

Are you requesting a change in the app's code or the test?

hueitan:
oh wait, zeplin has close button on the left, let's get clarify from design.

cypress/integration/spec.js
hueitan: I'm not very sure, let's resolve this.

PR #23, opened 2019-12-05
Show the last page when user goes to the previous section

hueitan:
`useLayoutEffect` works and fixes

> When looking at the section title page, the UP key goes to the first page of the previous section. The last page of the previous section would be more intuitive.

@stephanebisson do you have any idea how to redesign the relationship between Article, Pager, useScroll?


stephanebisson:
> @stephanebisson do you have any idea how to redesign the relationship between Article, Pager, useScroll?

We could get rid of Pager and do something like `const [currentPage, setCurrentPage, isFirstPage, isLastPage, etc?] = usePagination()` in <Article>. The goal is to encapsulate all the pagination behavior in one place.

hueitan:
Thanks for the suggestion!

`useScroll` stays until we touch the quickfact and see whether it is needed.

🎉 All the pagination behaviour is now under `usePagination()`

hueitan:
@stephanebisson can you also check whether this still exists after this change?

> On the device, I noticed that if I click down quickly after landing on a new section I can skip it entirely and go to the next section. From there, if I click up to go to the previous section, I can click slowly and see all the pages of the section that was skipped.

I don't get it from my device

src/hooks/usePagination.js
```diff
@@ -0,0 +1,63 @@
...
+
+  const showPrevPage = () => {
+    const prevPage = currentPage - 1
+    if (prevPage >= 0) {
+      setCurrentPage(prevPage)
+      setIsLastPage(true)
+    }
+  }
+
+  return [currentPage, isLastPage]
```
stephanebisson: Is `isLastPage` being used externally?

src/components/Article.js
```diff
@@ -82,15 +81,16 @@ const ArticleSection = ({
...
   const article = useArticle(lang, title)
   const softkey = useSoftkey()
-  const [currentSection, setCurrentSection] = useState(0)
 
   if (!article) {
     return 'Loading...'
   }
 
-  const section = article.sections[currentSection]
+  const [currentPage] = usePagination(containerRef, 240, 'x', article.sections.length)
```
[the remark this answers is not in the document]
hueitan: > We're not using the current page information just et but we will if we want to update the softkeys based on the actions (sections, quick facts, audio) presence.

when the title and description are long, then the action button will be shown on the second page, so checking the current page might not be work.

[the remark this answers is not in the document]
stephanebisson: That's another thing to fix; they need to be on the first page.

src/hooks/usePagination.js
```diff
@@ -0,0 +1,63 @@
+import { useState, useLayoutEffect } from 'preact/hooks'
+import { useKeys } from 'hooks'
+
+export const usePagination = (
+  elementRef,
+  step,
```
stephanebisson: Do we need `step` and `axis` arguments here?

src/hooks/usePagination.js
```diff
@@ -0,0 +1,63 @@
...
+  const prop = axis === 'x' ? 'scrollLeft' : 'scrollTop'
+
+  useKeys({
+    ArrowDown: () => {
+      const previous = elementRef.current[prop]
+      elementRef.current[prop] += step
+      const after = elementRef.current[prop]
+
+      // show the next section of the article
+      if (previous === after) {
```
stephanebisson: On the device, I can skip part of sections, this is the logic responsible.

stephanebisson: It was actually the CSS that was responsible. I reordered the DIVs and the classes and it fixed it for me. Please test.

hueitan: Confirmed, with this change, I can see it works for my firefox browser, unlike before it works only for first and second page when I paginate.

Can you explain your change?

hueitan:
The current solution of handling the sections and pages relationship all in the hooks `usePagination`, I would like to rename to `useArticlePagination` as this is only used by Article page. The benefit is we can find all the pagination related state within one hook, instead of the logic and bubble event across different view and hook.

In the future, if there is a new requirement that only handles page, then the code will be easier without consider section. For example, `quickfact` only has 1 section, so using `useScroll` should be enough.

src/hooks/usePagination.js
stephanebisson: The `page` class needs to be at the top. The name is misleading. It is not related to pagination. It predates it. It represents a page in the general sense, like the search page or the article page.

Article pagination needs to DIVs to work, the outer DIV with the `pages-container` class that handles overflow, and the inner DIV with the 'pages' class that has the column layout. The page class was interfering with the inner one, forcing it to actually be 480px wide. That value is there to trick Firefox, it's not meant to be the actual width.

PR #26, opened 2019-12-10
Add <Toc/> Component

hueitan:
<s>🚧 Don't merge</s>
---

Toc is a "popup" view component in Article page, then
- [x] Fix the key press conflict event in `useNavigation` `usePagination`
- [x] Close button in Toc page should go to the previous shown location
- [x] Enter button in Toc page should go to the specific header location
- [x] Fix the location of popup, given that pagination will change the scrollLeft location
- [x] Consider the case right key open menu first, then open section list


hueitan:
@stephanebisson you can test it with the right key to open Section list

src/hooks/useArticlePagination.js
```diff
@@ -5,14 +5,18 @@ const DEVICE_WIDTH = 240
...
 ) => {
-  const [currentSection, setCurrentSection] = useState(0)
+  const [currentSection, setCurrentSection] = useState(findLocatedSection(article.toc, subTitle))
   const [isLastPage, setIsLastPage] = useState(0)
   const prop = 'scrollLeft'
+  const numOfSection = article.sections.length
 
   useKeys({
     ArrowDown: () => {
+      if (showToc) return
```
stephanebisson: I think this line represents the main issue I see with this structure. We need to start thinking about better systems
* popups: so that their logic is not completely embedded in the components
* softkeys: to allow popups and other components to register (on show) and unregister (on hide) so we don't have to keep track of all the states in the handlers.

src/hooks/useArticlePagination.js
```diff
@@ -59,5 +85,17 @@ export const useArticlePagination = (
     }
   }
 
-  return [currentSection]
+  return [currentSection, setCurrentSection]
+}
+
+const findLocatedSection = (toc, title) => {
```
stephanebisson: Name could be simplified to `findSection`

src/api/getArticle.js
```diff
@@ -53,11 +54,27 @@ export const getArticle = (lang, title) => {
...
+
+      // build toc structure
+      switch (s.toclevel) {
+        case 1 :
+          toc.push(s.line)
+          break
+        case 2 : {
+          const index = toc.length - 1
+          if (typeof toc[index] === 'string') {
+            toc[index] = [toc[index]]
```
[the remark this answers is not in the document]
hueitan: Thanks for the suggestion @stephanebisson !! The entire data structure of toc looks easier now! Thanks alot!

PR #27, opened 2019-12-10
Select article links

stephanebisson:
Use left/right to cycle through article links and actions and enter to go to the page or invoke the action.


hueitan:
It looks good to me, the only thing is the left key "Close", it doesn't close and route to Search page when entering another article link.

PR #28, opened 2019-12-12
Refactor useSoftkey

stephanebisson:
* Make useSoftkey more declarative
* Allow popup to have their own softkeys while
  they are open and have the parent's component
  softkey continue working once the popups are
  closed.


src/hooks/useSoftkey.js
```diff
@@ -1,4 +1,21 @@
...
+    return () => { newContext && softkey.dispatch({ type: 'pop' }) }
+  }, [newContext])
+
+  useEffect(() => {
+    if (config) {
+      softkey.dispatch({ type: 'set', config })
+    }
+  }, dependencies)
+  return softkey
+}
```
hueitan: route to a new page, popup a new component, always `newContext=true`

in which case do you think the `newContext` needed to be `false` here?

stephanebisson: In this commit, it is `false` in `<ArticleSection>` to signify to it contributes to the context created by `<Article>` and doesn't replace it. 

Another way to achieve something similar would be to use a string like the `origin` param to specify which view a component belongs to. All components of the article view could specify `article` to share a context.

stephanebisson: The second option proposed here has the advantage of being independent from the call order.

hueitan: 👍 I see the latest commit, `origin` looks clearer to define which context it is.

src/hooks/useArticlePagination.js
```diff
@@ -1,22 +1,20 @@
...
 ) => {
   const [currentSection, setCurrentSection] = useState(findSection(article.toc, subTitle))
   const [isLastPage, setIsLastPage] = useState(0)
   const prop = 'scrollLeft'
   const numOfSection = article.sections.length
 
-  useKeys({
-    ArrowDown: () => {
-      if (isTocShown) return
+  useSoftkey('Article', {
```
hueitan: 👍👍👍 This is really awesome!

src/components/Article.js
```diff
@@ -14,30 +14,34 @@ const ArticleBody = memo(({ content }) => {
...
-      type: 'setCenterText',
-      value: hasActions ? i18n.i18n('centerkey-select') : ''
-    })
-    if (hasActions) {
-      setNavigation(0)
-    }
-  }, [hasActions])
-
+  const [,, getCurrent] = useNavigation('Article', actionsRef, 'x')
+  const [isVisible, setVisible] = useState(false)
```
hueitan: should this rename to which component is visible? or if this is the temporary state, then make the comment so we know it when we reach this again.

src/components/ArticleToc.js
```diff
@@ -6,19 +6,8 @@ import { ListView } from 'components'
 export const ArticleToc = ({ items, close }) => {
```
hueitan: I would also like to rename the close function to be `closeFn`, same as the audio popup, I will do it in my later bug fix pr.

src/components/Search.js
```diff
@@ -5,35 +5,28 @@ import { useNavigation, useSearch, useLanguage, useI18n, useSoftkey } from 'hook
...
-
   const onKeyCenter = () => {
     const { index, key } = getCurrent()
     if (index) {
       window.location.hash = `/article/${lang}/${key}`
     }
   }
 
+  useSoftkey('Search', {
+    left: 'Settings',
```
hueitan: `i18n.i18n( 'settings' )`

src/components/Article.js
stephanebisson: Done

PR #29, opened 2019-12-13
Fix ToC bug found

hueitan:
let me know if anyone found more issues on ToC

### DIFF
--- a/src/hooks/useArticlePagination.js
+++ b/src/hooks/useArticlePagination.js
@@ -58,13 +58,16 @@ export const useArticlePagination = (
 
       subTitleElement && subTitleElement.scrollIntoView()
 
-      // @todo replace the magic number with constant device width
-      if (elementRef.current.scrollLeft % 240 === 216) {
-        elementRef.current.scrollLeft += 24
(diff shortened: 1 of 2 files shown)


src/hooks/useArticlePagination.js
```diff
@@ -58,13 +58,16 @@ export const useArticlePagination = (
 
       subTitleElement && subTitleElement.scrollIntoView()
 
-      // @todo replace the magic number with constant device width
-      if (elementRef.current.scrollLeft % 240 === 216) {
-        elementRef.current.scrollLeft += 24
+      elementRef.current.scrollTop = 0
+      // @todo replace the DEVICE_WIDTH to utils
+      // @todo need to verify the magic number again when testing different device width
+      if (elementRef.current.scrollLeft % DEVICE_WIDTH === 216) {
```
stephanebisson: I'm going to merge this PR but I would like us to rewrite this part without using this magic number.

We should be able to locate the target element, figure out on which page it is based on its offset and go directly there without using `scrollIntoView()` with some corrections.

PR #35, opened 2019-12-16
Generic popup component

stephanebisson:
This is the beginning of a generic popup component. 

TODO
- [x] Move popup styling from ArticlePreview to popup
- [x] Respect popup options (position: 'bottom' or 'fullscreen')

To be done in followup PRs
- Use it for the TOC and maybe QuickFacts
- Support several popups on top of one another (a menu that opens up another menu)


hueitan:
:+1: Looks good to me

PR #36, opened 2019-12-17
Support Responsive Layout

hueitan:
I replace most of them using % so that it can fit _"most"_ of the device.

I also update some stylings as shown below 

Screen Shot 2019-12-17 at 10 58 22 AM
_(highlight selected a text)_

Screen Shot 2019-12-17 at 10 58 16 AM
_(highlight selected action button)_

### DIFF
--- a/style/layout.less
+++ b/style/layout.less
@@ -9,9 +9,5 @@ body {
     #root {
         padding: 0;
         margin: 0;
-
-        .page {
-            // @todo confirm whether we need the page class name
-        }
     }
 }
--- a/style/layout.less
+++ b/style/layout.less
@@ -7,13 +7,11 @@ body {
     margin: 0;
 
     #root {
-        width: 100%;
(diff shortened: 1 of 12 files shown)


hueitan:
focus on testing with 240x320 device, we need to also check 320x240 but not available right now.

style/layout.less
```diff
@@ -7,13 +7,11 @@ body {
...
 
     #root {
-        width: 100%;
         padding: 0;
         margin: 0;
 
         .page {
-            width: 100%;
-            height: @pageHeight;
+            // @todo confirm whether we need the page class name
```
stephanebisson: The idea was to specify the fullscreen nature of each top-level components (Search, Article, etc) here instead of individually.

If it's not needed anymore, feel free to remove it.

hueitan: Done.

PR #41, opened 2019-12-19
Let the language popup adjust to content size

stephanebisson:


stephanebisson:
Oups, nevermind. The support for 'auto' is in my current reference preview branch.

PR #42, opened 2019-12-19
Reference preview

stephanebisson:


stephanebisson:
Notes are similar to references but are mostly unstructured in the article. They won't be supported in this PR.

src/components/ReferencePreview.js
```diff
@@ -0,0 +1,20 @@
+import { h } from 'preact'
+import { useRef } from 'preact/hooks'
+import { useI18n, useSoftkey, useArticleLinksNavigation } from 'hooks'
+
+export const ReferencePreview = ({ reference, lang, close }) => {
+  const i18n = useI18n()
+  const contentRef = useRef()
+  useArticleLinksNavigation('ReferencePreview', lang, contentRef, 1)
+  useSoftkey('ReferencePreview', {
+    right: i18n.i18n('close'),
```
hueitan: `softkey-close`

src/hooks/useArticleLinksNavigation.js
```diff
@@ -1,18 +1,30 @@
...
   currentPage,
-  onTitleClick,
-  onActionClick
+  linkHandlers = {}
 ) => {
+  const i18n = useI18n()
   const [links, setLinks] = useState([])
   const [currentLink, setCurrentLinkInternal] = useState(null)
 
+  const [showArticlePreview] = usePopup(ArticlePreview, { position: 'bottom' })
```
hueitan: sometimes the preview only have few lines, then we will have a lot of blank content. do you think 'auto' here better? 

but then the 'loading' effect feel bad =( (or we can do the same as [loading article](https://phabricator.wikimedia.org/T240888), less than 500ms don't show loading)
Screen Shot 2019-12-21 at 12 14 33 AM

src/components/ReferencePreview.js
stephanebisson: Done

src/hooks/useArticleLinksNavigation.js
stephanebisson: Yeah, it turns out 'auto' is the only mode that makes sense. I'll simplify this.

stephanebisson:
I don't know what you mean for `en/Catholic Church` ref 4. I can select all 3 links.
Screen Shot 2019-12-20 at 7 52 47 PM

src/hooks/useArticleLinksNavigation.js
stephanebisson: In our target audience, loading is unlikely to be fast. Whatever we do for the main article loading screen, we should do something similar here.

hueitan:
I just tried on the `ext link`, on browser press `enter/return` key doesn't open the new link, but it does when clicking the `SELECT` key

stephanebisson:
We use `window.open` for external links. Is it possible your browser blocks those? You should try on a device to be sure.

hueitan:
confirm and merge. let's open a separate issue when we found something.

PR #43, opened 2019-12-19
usePopup for article page toc component

hueitan:


src/components/Article.js
```diff
@@ -81,16 +81,16 @@ const ArticleInner = ({ lang, articleTitle, initialSubTitle }) => {
     return 'Loading...'
   }
 
-  const [isTocShown, toggleToc] = useState(false)
   const [subTitle, setSubTitle] = useState(initialSubTitle)
+  const [popupToc] = usePopup(ArticleToc)
```
stephanebisson: For consistency, the function here should be named show<Something>Popup(). For instance `showArticleTocPopup` or just `showTocPopup`

src/components/ArticleToc.js
```diff
@@ -3,7 +3,7 @@ import { useRef, useEffect } from 'preact/hooks'
 import { useNavigation, useI18n, useSoftkey } from 'hooks'
 import { ListView } from 'components'
 
-export const ArticleToc = ({ items, close }) => {
+export const ArticleToc = ({ items, closeFn, close }) => {
```
stephanebisson: We can probably find a better name for `closeFn`. Maybe `onSelectItem`...

src/components/Article.js
hueitan: see if it is better now

style/toc.less
```diff
@@ -13,7 +13,7 @@
 
 	.list {
 		top: @pageHeaderHeight;
-		bottom: @softkeysHeight;
+		bottom: 0px;
```
stephanebisson: 0 should be just 0 with no units.

PR #44, opened 2019-12-19
the close key should always on right softkey

hueitan:
quick update of https://phabricator.wikimedia.org/T241044

so we can feel the consistency with the physical device keyboard ^^


hueitan:
ok, test spec needed to update

hueitan:
Master merged. This is also ready to be reviewed now.

PR #46, opened 2019-12-20
Don't justify article text

stephanebisson:
I don't know why I did that. It's not in the spec and it's ugly.


hueitan:
without justify
Screen Shot 2019-12-20 at 3 55 22 PM

with justify
Screen Shot 2019-12-20 at 3 55 17 PM

I personally prefer to justify it

stephanebisson:
See what it does to links:
Screen Shot 2019-12-20 at 10 03 29

PR #48, opened 2019-12-20
Change "Language setting" to "Language settings"

amire80:
More consistent with other apps, and that's also what
the message documentation says.

Feel free to close it if it is supposed to say "Language setting".


stephanebisson:
Makes sense to me but I'll let the designer and pm make the call. I've asked on phab.

PR #51, opened 2019-12-21
Encapsulate MW API URL in utils

stephanebisson:
`URLSearchParams` doesn't appear to be working correctly in FF48. This change at least encapsulates the MW url building in one place.

### DIFF
--- a/src/utils/mwApi.js
+++ b/src/utils/mwApi.js
@@ -1,0 +1,13 @@
+const defautParams = {
+  format: 'json',
+  formatversion: 2,
+  origin: '*'
+}
+
+export const buildMwApiUrl = (lang, params) => {
+  params = Object.assign({}, defautParams, params)
+  const baseUrl = `https://${lang}.wikipedia.org/w/api.php`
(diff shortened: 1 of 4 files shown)


hueitan:
it's weird 😦 the Browser compatibility said it supports it starting ver `29`, anyway I'm going to merge this, nothing serious to block us.

src/utils/mwApi.js
```diff
@@ -0,0 +1,13 @@
...
+  format: 'json',
+  formatversion: 2,
+  origin: '*'
+}
+
+export const buildMwApiUrl = (lang, params) => {
+  params = Object.assign({}, defautParams, params)
+  const baseUrl = `https://${lang}.wikipedia.org/w/api.php`
+  return baseUrl + '?' + Object.keys(params).map(p => {
+    return `${p}=${encodeURIComponent(params[p])}`
```
hueitan: sorry I think I merged too fast without checking this properly

because the encoding here, if you search for `wikimedia foundation`, you won't get the result from api.

the api doesn't treat `%20` as a space.

PR #58, opened 2019-12-23
Change Article Language Feature

hueitan:
screenshot
Screen Shot 2019-12-23 at 10 38 35 PM


hueitan:
The `change article language` code and behaviour look the same as `change app language`

but there are some differences :
1. initial list (fetch vs local var)
1. header title
1. listview list item includes / not include current selected item
1. one used in page component, another used in popup component
1. one change the global i18n.locale, another change only the route `lang`

we can think of using the same hook same view, but I would like to have the team to play with both language change behavior, and see whether we need any changes to them

src/hooks/useSearchArticleLanguage.js
```diff
@@ -0,0 +1,59 @@
...
+
+export const useSearchArticleLanguage = (lang, title) => {
+  const [items, setItems] = useState([])
+  const [query, setQuery] = useState()
+  const [allLanguages, setAllLanguages] = useState([])
+
+  useEffect(() => {
+    getLanglinks(lang, title)
+      .then(items => {
+        const languages = items.map(item => {
```
stephanebisson: This map could be in the API transform function to do it only once and cache it.

hueitan: good point! on it now.

PR #62, opened 2019-12-27
<TextSize/> Component & Adjust Feature

hueitan:
<s>_still missing the text size adjust feature_</s>


hueitan:
For the second part, does this pr helps? https://github.com/wikimedia/wikipedia-kaios/pull/67

stephanebisson:
> For the second part, does this pr helps? #67

Yes, I think it would work if we also pass in the text size in the dependencies array.

src/utils/articleTextSize.js
```diff
@@ -0,0 +1,66 @@
...
+  const onKeyboard4 = () => adjust(1)
+  const onKeyboard5 = () => reset()
+  const onKeyboard6 = () => adjust(-1)
+
+  return {
+    onKeyboard4, onKeyboard5, onKeyboard6
+  }
+}
+
+export default { get, set, adjust, reset, init, getSoftkeyEffect }
```
stephanebisson: Please use a named export so the name is maintained here and not in the index file.

PR #63, opened 2019-12-30
Add anchor goto confirm dialog in article link

hueitan:
fixed https://phabricator.wikimedia.org/T241364

styling pending

Annotation 2019-12-31 214625


src/components/Article.js
```diff
@@ -35,6 +37,10 @@ const ArticleSection = ({
     },
     reference: ({ referenceId }) => {
       showReferencePreview({ reference: references[referenceId], lang })
+    },
+    goto: ({ anchor }) => {
```
stephanebisson: I don't think `goto` is appropriate. The other keys represent what the link is. Maybe `section` would be more clear.

src/components/Article.js
```diff
@@ -35,6 +37,10 @@ const ArticleSection = ({
...
+      // @todo styling to be confirmed with design
+      confirmDialog({ message: `Go to Section "${anchor}"?`, onSubmit: () => goToSubpage({ title: anchor }) })
```
stephanebisson: This message needs to be localized.

src/utils/confirm.js
```diff
@@ -0,0 +1,7 @@
+import { usePopup } from 'hooks'
+import { ConfirmDialog } from 'components'
+
+export default function ({ message, onSubmit }) {
```
stephanebisson: Is there a strong reason to have a different style here? All the components are named export of a lambda and this is a default export of a function.

src/components/ConfirmDialog.js
```diff
@@ -0,0 +1,23 @@
...
+  useSoftkey('ConfirmDialog', {
+    right: i18n.i18n('softkey-close'),
+    onKeyRight: close,
+    center: i18n.i18n('softkey-ok'),
+    onKeyCenter: () => { onSubmit(); close() }
+  }, [])
+
+  return (
+    <div class='confirm-dialog'>
+      <div class='item'>
```
stephanebisson: Why do we have this hierarchy or `.item > .info > .title` to display one string?

src/utils/confirm.js
hueitan: true, this just returns a single function.

src/hooks/useArticleLinksNavigation.js
```diff
@@ -106,15 +107,18 @@ const makeLinkClickEvent = link => {
...
+    const referenceId = link.getAttribute('href').slice(1)
     return { type: 'reference', referenceId }
   }
 
   if (link.getAttribute('rel') === 'mw:ExtLink') {
     return { type: 'external', href: link.href }
   }
+
+  if (link.hash) {
+    return { type: 'section', anchor: link.textContent }
```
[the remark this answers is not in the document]
hueitan: good catch! text and anchor might be different!

src/components/ConfirmDialog.js
hueitan: good catch, i was copying this from listview, and we don't need item as the class name!

I also make the change to fit longer section name
Capture

i18n/qqq.json
```diff
@@ -10,6 +10,7 @@
 	"app-title": "Title of the application. Shown in the app store and the app launcher.",
 	"app-subtitle": "Subtitle of the application. Shown in the app store.",
 	"app-description": "Description of the application. Shown in the app store.",
+	"confirm-section": "Confirm message switch to the selected section page. Shown in the Article page.",
```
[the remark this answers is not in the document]
hueitan: done.

PR #65, opened 2019-12-31
Requesting mw api from mobile api

hueitan:
when playing with mobile web page and the footer section, I found sometimes the footer will break due to the request is not trigger from mobile version, then I found all the request in mobile web page are `<lang>.m.wikipedia.org/w/api.php`

maybe can someone tell the different? and should we use `m` for all the request?

### DIFF
--- a/src/utils/mwApi.js
+++ b/src/utils/mwApi.js
@@ -6,7 +6,7 @@ const defautParams = {
 
 export const buildMwApiUrl = (lang, params) => {
   params = Object.assign({}, defautParams, params)
-  const baseUrl = `https://${lang}.wikipedia.org/w/api.php`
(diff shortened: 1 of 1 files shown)


hueitan:
I was trying this before, for the article footer
https://en.wikipedia.org/w/api.php?format=json&formatversion=2&origin=*&action=query&prop=pageimages|description&piprop=thumbnail&pithumbsize=160&pilimit=3&generator=search&gsrsearch=morelike:Carnivorous&gsrnamespace=0&gsrlimit=3&gsrqiprofile=classic_noboostlinks&uselang=content

but I can't reproduce it now, is there any different from the API side?

stephanebisson:
While it's technically possible for API modules to behave differently in mobile mode, I've never seen it done in practice. Let's not use the `.m` urls.

PR #64, opened 2019-12-31
Add <ArticleFooter/>

hueitan:
more to be done

- [x] i18n footer content
- [x] footer title in toc
- [x] thumbnail selected styling
- [x] highlight first item when the page load (article link)

PR #67, opened 2020-01-02
Minor change to useArticleLinksNavigation

hueitan:
the change is because of the article footer

1. add dependencies as the array to allow more state (page:String -> dependencies:[])
2. return `findAndSelectCurrentLink()`, useful for the page that has ajax request (eg: footer page)

please find article footer branch to checkout the usage of `findAndSelectCurrentLink()`

### DIFF
--- a/src/components/Article.js
+++ b/src/components/Article.js
@@ -42,7 +42,7 @@ const ArticleSection = ({
     }
   }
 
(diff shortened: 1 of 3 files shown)


hueitan:
doesn't need anymore because of the change here
https://github.com/wikimedia/wikipedia-kaios/pull/64/commits/1c4a97299b3245f2ce06b8f9bfd0f01afc7a49ce

PR #68, opened 2020-01-03
add test to change language

jpita:


cypress/integration/settings-spec.js
```diff
@@ -33,9 +36,28 @@ describe('settings page', () =>{
   })
 
   it('only the first element should be selected on load', () => {
-    firstElementOfTheSettingsMenuList.should('have.attr', 'nav-selected', 'true')
```
jpita: for some reason this "should" was not asserting anything, I couldn't use firstElementOfTheSettingsMenuList here but on the next line it works.
if anyone can find the "why" that would be great.

hueitan: it works from my side, does it work as well in CI?

jpita: the test passes but the assert is never done.
you can see by running `npx cypress open` on master branch and run that test.
there is no asserts being run.
image

cypress/integration/settings-spec.js
```diff
@@ -3,9 +3,11 @@
 
 import { SettingsPage } from "../page-objects/settings-page"
 import { SearchPage } from "../page-objects/search-page"
+import { LanguageSettingsPage } from "../page-objects/language-settings-page"
```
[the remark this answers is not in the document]
jpita: are they not different?
if they're different makes sense to have 2 (both in src and tests), if they're the same (in terms of selectors) we can have just one in the tests

cypress/integration/settings-spec.js
```diff
@@ -17,7 +19,8 @@ describe('settings page', () =>{
   })
 
   it('show all the items in the list', () => {
-    settingsPage.settingsList().should('have.text', "LanguageText sizeAbout WikipediaPrivacy policyTerms of useRate the appHelp and feedbackAbout the app")
+    settingsPage.settingsList().should('have.text', 
+    "LanguageText sizeAbout WikipediaPrivacy policyTerms of useRate the appHelp and feedbackAbout the app")
```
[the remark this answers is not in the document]
jpita: I prefer your idea of getting the content from the json file but I don't know how to do that in javascript.
can you help me with that (will be fast) or should I look it up (will probably take a while)?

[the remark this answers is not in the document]
hueitan: try `import * as data from '../../i18n/en.json'` in this file

cypress/integration/settings-spec.js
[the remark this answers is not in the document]
hueitan: it looks the same but there are differences, [see the differences here](https://github.com/wikimedia/wikipedia-kaios/pull/58#issuecomment-568592252)

stephanebisson:
> Do you think it's better to read the `i18n/en.json` file for all the text? Otherwise whatever translatewiki change will cause the CI failed.

`en.json` is managed by us in the code. TWN won't change it.

cypress/page-objects/language-settings-page.js
```diff
@@ -0,0 +1,13 @@
...
+    }
+
+    popupTitleElement(){
+        return cy.get('.language-message > .header')
+    }
+
+    headerElement(){
+        return cy.get('.header')
+    }
+}
```
stephanebisson: Missing a line break at the end of this file.

jpita: fixed

hueitan:
true, `en.json` is mantained by us. but anyway, getting the text from `en.json` still recommended 😃

cypress/integration/settings-spec.js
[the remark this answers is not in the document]
jpita: I prefer to have `LanguageSettingsPage` and `LanguageArticlePage` .
Is that ok with you?

cypress/integration/settings-spec.js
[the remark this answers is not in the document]
jpita: @hueitan I added the code to use the json files for the strings, can you check now?

cypress/integration/settings-spec.js
```diff
@@ -15,7 +37,7 @@ describe('settings page', () => {
   })
 
   it('show all the items in the list', () => {
-    settingsPage.settingsList().should('have.text', 'LanguageText sizeAbout WikipediaPrivacy policyTerms of useRate the appHelp and feedbackAbout the app')
+    settingsPage.settingsList().should('have.text', settingsMenuListEnglishText.toString().replace(/,/g, ''))
```
stephanebisson: To make a string of all the elements in the array joined together, I recommend using `settingsMenuListEnglishText.join('')`

cypress/integration/settings-spec.js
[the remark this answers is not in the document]
hueitan: Looks good now

cypress/integration/settings-spec.js
```diff
@@ -31,8 +53,26 @@ describe('settings page', () => {
...
   })
+
+  it('language of the app should change', () => {
+    cy.enter()
+    languageSettingsPage.popupTitleElement().should('have.text', enJson['language-setting'])
+
+    languageSettingsPage.popupTextElement().should('have.text', languageSettingsPopupEnglishText)
+    cy.enter()
+
+    cy.get('.left').should('have.text', 'Search')
```
hueitan: 'Search' here use `enJson['softkey-search']`

hueitan: consider making Softkey also in the page-object, then you can `softkey.leftText()` or just `.left()`

jpita: fixed

cypress/integration/settings-spec.js
jpita: fixed

cypress/integration/settings-spec.js
```diff
@@ -37,7 +37,8 @@ describe('settings page', () => {
   })
 
   it('show all the items in the list', () => {
-    settingsPage.settingsList().should('have.text', settingsMenuListEnglishText.toString().replace(/,/g, ''))
+    settingsPage.settingsList().should('have.text', settingsMenuListEnglishText.join(''))
+    //settingsMenuListEnglishText.join('')
```
stephanebisson: Please remove commented out code

jpita: oops

jpita: fixed

cypress/support/commands.js
```diff
@@ -43,6 +43,15 @@ Cypress.Commands.add('clickCloseButton', () => {
 Cypress.Commands.add('clickSettingsButton', () => {
   cy.get('label.left').contains('Settings').click()
 })
+
 Cypress.Commands.add('navigateToHomePage', () => {
   cy.visit('http://127.0.0.1:8080')
 })
+
+Cypress.Commands.add('getLeftSoftkeyButton', () => {
+  return cy.get('.left')
```
hueitan: I recommend `cy.get( '.softkey > .left' )` here to get the specific softkey left element, as `.left` may possibly exist in other component

jpita: fixed and added another small improvement, does it look ok?

PR #70, opened 2020-01-06
Article Loading Experience

hueitan:


src/components/Loading.js
```diff
@@ -0,0 +1,25 @@
...
+      history.back()
+    }
+  }
+
+  useSoftkey('Loading', {
+    right: i18n.i18n('softkey-close'),
+    onKeyRight: onClose
+  })
+
+  return <div class='loading-planet'>
```
stephanebisson: There should be parenthesis around the JSX here.

src/components/Loading.js
```diff
@@ -0,0 +1,25 @@
+import { h } from 'preact'
+import { useSoftkey, useI18n } from 'hooks'
+export const Loading = ({ close, message }) => {
+  const i18n = useI18n()
+
+  const onClose = () => {
+    if (close) {
```
stephanebisson: Is the `close` property used anywhere?

hueitan: I'm thinking of `<Loading/>` can be used either as a single page or popup content, but yes we are only using this in article so far, so I will remove this line until we need it.

src/components/Loading.js
hueitan: I'm not sure what it means here, but this returns a single <div/>, that's fine wihout having the parenthesis here, is there any code styling concern here?

stephanebisson: This is for consistency with the other components. Put `(` right after `return` and start the JSX on the next line. This way the markup is indented correctly.

PR #79, opened 2020-01-10
Translation of the article actions button text

hueitan:
https://phabricator.wikimedia.org/T234628#5792612


i18n/en.json
```diff
@@ -6,6 +6,8 @@
         "locale": "en",
         "message-documentation": "qqq"
     },
+    "article-action-sections": "Sections",
```
stephanebisson: We need the menu item and article action button to use the same message. I think `article-action-` is a good prefix for that even when shown in the article menu. The qqq should say which features it triggers and that it can be displayed on the page or in a menu.

We should also cleanup `softkey-sections` if it's not used anymore.

hueitan: Done.

PR #84, opened 2020-01-13
Onboarding Component

hueitan:


hueitan:
@jpita do you want to work on adding the localStorage to skip the onboading page? so the cypress test can pass.

src/components/Onboarding.js
```diff
@@ -0,0 +1,37 @@
+import { h } from 'preact'
+import { useState, useEffect } from 'preact/hooks'
+import { route } from 'preact-router'
+import { useSoftkey, useI18n } from 'hooks'
+import { onboarding } from 'utils'
+
+export const Onboarding = () => {
```
stephanebisson: I would like to see this component simplified. There is a lot of duplication and a lot of calls to i18n() on each render for the same keys.

src/components/Onboarding.js
```diff
@@ -0,0 +1,37 @@
...
+  const i18n = useI18n()
+  const [currentIndex, setCurrentIndex] = useState(0)
+  const softkey = useSoftkey('origin')
```
stephanebisson: I'm curious why the regular `useSoftkey('origin', { config })` didn't work here

hueitan: oh, it should be `onboarding` not `origin`

`useSoftkey('onboarding', { config })` works, `config` needs to be dynamic when `currentIndex ` change, our current hook doen't allow that

src/components/Onboarding.js
hueitan: we can also compile the output in `object { image, background, title, description }`, and listen to `currentIndex`, but this brings extra state which doing the same thing.

src/components/Onboarding.js
stephanebisson: `useSoftkey()` has a `dependencies` argument that could contain `currentIndex`. I don't know if it would work or not.

hueitan: it doesn't work, the current solution works only for function call in config `onKeyLeft` `onKeyRight` `onKeyCenter`, `left` `right` are text value that not changing.

hueitan:
@stephanebisson can we let this go?

src/components/Onboarding.js
stephanebisson: I'll make it work when I have time.

stephanebisson:
@jpita you can merge this when you're ready

jpita:
@stephanebisson can I move the task to done as well?

src/components/Onboarding.js
hueitan: The follow up pr for this is here https://github.com/wikimedia/wikipedia-kaios/pull/92

PR #85, opened 2020-01-14
 Use summary API when preview article

hueitan:
Phab. : https://phabricator.wikimedia.org/T242082


src/api/getArticleSumary.js
```diff
@@ -0,0 +1,11 @@
+import { cachedFetch } from 'utils'
+
+export const getArticleSummary = (lang, title) => {
+  const url = `https://${lang}.wikipedia.org/api/rest_v1/page/summary/${encodeURIComponent(title)}`
```
hueitan: Let's do the abstraction when we modify `getArticle` and use new endpoint.

hueitan:
@stephanebisson can you show me the example of `type: 'no-extract'`, I can't find one.

stephanebisson:
On `en/Cat` the "carnivorous" link in the first paragraph returns "no-extract" for me. Let me know if you are seeing the same thing.

hueitan:
Pending : Add loading experience when fetching the API.

hueitan:
Capture

as agree with the design, we show the title when loading so user knows which article is being fetched.

hueitan:
🐛 need to update the `text size adjustment` here .. argz,,,

hueitan:
alright, this is ready to be reviewed again 🔥

jpita:
ah damn, did `git push` instead of `git pull`.
can you check if I need to revert the commit @hueitan ?
it seems that it just brought the changes from master

hueitan:
No, that’s fine, you don’t need to revert it.

jpita:
> @jpita What's the status here, can we merge it soon?

sorry, totally forgot about this due to the task not being in QA

PR #90, opened 2020-01-15
Onboarding tests

jpita:
🚨 DEPENDS ON #84, MERGE THIS ONLY AFTER #84 IS MERGED ⚠️ 

* adding tests for onboarding 
* adding package [cypress-localstorage-commands](https://github.com/javierbrea/cypress-localstorage-commands) to handle localstorage variables with cypress (I haven't checked it for security issues, it's not critical, we can remove it)


jpita:
I'm not checking for the background image because it's missing on page 0 on #84 @hueitan .
Once fixed I'll add the check to all the pages

jpita:
> Missing `Get Started` Test and `enJson` for softkey text

fixed

hueitan:
Missing `Get Started` and load Search page test, similar to `Skip` in your test case.

jpita:
> Missing `Get Started` and load Search page test, similar to `Skip` in your test case.

@hueitan if it's all fixed lmk so I can merge it

PR #92, opened 2020-01-18
Add more options to useSoftkey configuration

hueitan:
_Internal change how we pass the config to softkey component_

1. allow `replace` the softkey type, default to `set`

_Edit_: Turns out, get function is not needed

### DIFF
--- a/src/components/Onboarding.js
+++ b/src/components/Onboarding.js
@@ -1,5 +1,5 @@
 import { h } from 'preact'
-import { useState, useEffect } from 'preact/hooks'
+import { useState } from 'preact/hooks'
 import { route } from 'preact-router'
 import { useSoftkey, useI18n } from 'hooks'
 import { onboarding } from 'utils'
@@ -7,7 +7,6 @@ import { onboarding } from 'utils'
 export const Onboarding = () => {
   const i18n = useI18n()
   const [currentIndex, setCurrentIndex] = useState(0)
-  const softkey = useSoftkey('onboarding')
 
   const exitOnboard = () => {
     onboarding.markAsDone()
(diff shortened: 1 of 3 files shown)


src/components/Onboarding.js
```diff
@@ -33,10 +32,7 @@ export const Onboarding = () => {
     { left: i18n.i18n('softkey-back'), onKeyLeft: prevOnboard, right: i18n.i18n('softkey-next'), onKeyRight: nextOnboard },
     { left: i18n.i18n('softkey-back'), onKeyLeft: prevOnboard, center: i18n.i18n('softkey-get-started'), onKeyCenter: exitOnboard }
   ]
-
-  useEffect(() => {
-    softkey.dispatch({ type: 'replace', config: softkeyConfig[currentIndex] })
-  }, [currentIndex])
+  useSoftkey('onboarding', { config: softkeyConfig[currentIndex], replace: true }, [currentIndex])
```
stephanebisson: Have you tried adding all the keys (for example: `center: ''` to blank the center key) instead of adding the "replace" option?

hueitan: use still able to trigger the function without seeing the center text.

it works if we have `{ center: '', onKeyCenter: () => {}}`, but then we will need to add a lot of empty key to clear the key ☹️

stephanebisson: This is how it is used in places where softkeys are conditional.
`center: currentLink ? i18n.i18n('centerkey-select') : '',`

I don't mind this change but it makes the configuration more complex for a case that is not too common. Could we maybe add a 4th argument that defaults to the old behavior? Something like `action='set'`

hueitan: for the conditional softkey config, we will have

```js
const softkeyConfig = {
    left: currentIndex > 0 ? i18n.i18n('softkey-back') : i18n.i18n('softkey-skip'),
    onKeyLeft: currentIndex > 0 ? prevOnboard : exitOnboard,
    center: currentIndex === 3 ? i18n.i18n('softkey-get-started') : '',
    onKeyCenter: currentIndex === 3 ? exitOnboard : () => {},
    right: currentIndex === 3 ? '' : i18n.i18n('softkey-next'),
    onKeyRight: currentIndex === 3 ? () => {} : nextOnboard
  }
```

it looks even more complex, I will go for the 4th argument, adding the action key

PR #93, opened 2020-01-20
Special pages test

jpita:


cypress/page-objects/article-preview-page.js
```diff
@@ -0,0 +1,5 @@
+export class ArticlePreviewPage {
+  getTitle () {
+    return cy.get('.article-preview .title')
+  }
+}
```
hueitan: you define this but didn't use it?

cypress/integration/special-pages-spec.js
```diff
@@ -0,0 +1,17 @@
...
+  beforeEach(() => {
+    cy.navigateToHomePage()
+  })
+
+  it('special page article should open', () => {
+    cy.visit('http://127.0.0.1:8080/#/article/en/Help%3AIPA%2FEnglish')
+    articlePage.title().should('have.text', 'Help:IPA/English')
+    cy.get('.article-content').should('contain.text', 'Throughout Wikipedia, the pronunciation of words is indicated by means of the ')
+  })
+})
```
hueitan: not sure how special it is? are you going to test that opening the preview message, then goes to this article page?

cypress/page-objects/article-preview-page.js
jpita: will use it when the pagination works on cypress, right now I can't select the link I want to open on the preview

cypress/integration/special-pages-spec.js
jpita: right now I can't select the link I want to open on the preview , will complement the test when I can

PR #94, opened 2020-01-20
Support links navigation on quick facts page

stephanebisson:


stephanebisson:
More PRs to follow to complete the feature.

PR #95, opened 2020-01-20
Test search languages on settings and article

jpita:


cypress/integration/article-view-spec.js
```diff
@@ -0,0 +1,32 @@
...
+  beforeEach(() => {
+    cy.navigateToHomePage()
+  })
+
+  it('change article language', () => {
+    searchPage.search('cat')
+    searchPage.results().first()
+    cy.enter()
+    cy.downArrow()
+    cy.enter()
```
hueitan: just another though, do you want to support any one of these? so we don't end up creating a list of typing action.

````
cy.type( 'enter down enter' )
cy.type( [ 'enter', 'down', 'enter' ] )
cy.enter().downArrow().enter()
```

hueitan: just tried it, the third option already valid, i suggest use it.

jpita: fixed in all the tests

PR #97, opened 2020-01-20
Read canonical title when it's available in article preview

hueitan:
to avoid redirection, we get the actual title 

see https://phabricator.wikimedia.org/T242350#5816578

### DIFF
--- a/src/api/getArticleSummary.js
+++ b/src/api/getArticleSummary.js
@@ -3,7 +3,7 @@ import { cachedFetch, buildPcsUrl } from 'utils'
 export const getArticleSummary = (lang, title) => {
   const url = buildPcsUrl(lang, title, 'summary')
   return cachedFetch(url, data => ({
-    title: data.displaytitle,
+    titles: data.titles,
     imageUrl: data.thumbnail && data.thumbnail.source,
     preview: data.extract_html || data.description
   })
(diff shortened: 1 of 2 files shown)


stephanebisson:
Please provide an example of an API url that, when fetched using CORS, doesn't redirect correctly but provides the right canonical title.

hueitan:
load the article `Fassala`, when fetching the summary title `Hodh_Ech Chargui` `https://en.wikipedia.org/api/rest_v1/page/summary/Hodh_Ech Chargui`, it returns the canonical title in `Hodh_Ech_Chargui_Region`, so we know this is the title used to fetch article instead of `Hodh_Ech Chargui`

hueitan:
@stephanebisson not sure whether you can see the different, as you are in the region that api redirect doesn't always work correctly.

stephanebisson:
If you fetch "Hodh_Ech Chargui" and the canonical field contains "Hodh_Ech Chargui_Region", it is because the redirect worked for you and the URL you fetched is actually ".../Hodh_Ech Chargui_Region". 

The redirect issue is not really region-specific as other people have seen it from other places.

I'll merge this now. It doesn't resolve the main issue but it helps propagate the new title when the redirect does work.

PR #98, opened 2020-01-21
Add empty result found test case in search page

hueitan:
I test it all the time to prevent bad thing happens, so I'm adding one test here

### DIFF
--- a/cypress/integration/search-spec.js
+++ b/cypress/integration/search-spec.js
@@ -50,7 +50,7 @@ describe('Article search', () => {
       .children().first().next().should('have.class', 'img')
   })
 
-  it.only('should show empty result found when search for "adf23cv1"', () => {
+  it('should show empty result found when search for "adf23cv1"', () => {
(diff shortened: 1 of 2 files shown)


cypress/integration/search-spec.js
```diff
@@ -48,4 +49,19 @@ describe('Article search', () => {
...
+  it('should show empty result found when search for "adf23cv1"', () => {
+    searchPage.search('adf23cv1')
+    searchPage.getEmptyContent().should('have.text', enJson['no-result-found'])
+  })
+
+  it('center softkey should change when focus on result and search input', () => {
+    searchPage.search('cat')
+    cy.getCenterSoftkeyButton().should('not.have.text')
+    searchPage.results().first()
+    cy.downArrow().downArrow()
```
jpita: why `downArrow()` twice? any specific case ur looking for ? once works as well

hueitan: no specific reason for this, just some random test

PR #99, opened 2020-01-22
Quick Facts

stephanebisson:
* Constrain the width for most cases
* Move to a popup instead of a route
* Add article menu entry
* Allow changing text size

https://phabricator.wikimedia.org/T234619

Known issue: Opening and closing a preview from the Quick Facts popup, leads you back to the article, not the quick facts. This is because we don't yet support multiple popups on top of one another. This will be addressed in a subsequent PR.


hueitan:
We need to remove `goto.quickfacts` in goto utils.

PR #102, opened 2020-01-23
Support multiple popups

stephanebisson:
This is useful for situation where a popup brings another popup and closing the second one should return to the first one. For instance, showing article previews on top of quick facts.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -3,5 +3,6 @@
     "viewportWidth": 240,
     "viewportHeight": 290,
     "videoUploadOnPasses":false,
+    "modifyObstructiveCode": false,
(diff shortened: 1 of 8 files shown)


hueitan:
Just tried with the master, when having the second popup (article preview) in quickfacts, then click the center key, it does load the new article and close the second popup, but not the first popup.

stephanebisson:
> Just tried with the master, when having the second popup (article preview) in quickfacts, then click the center key, it does load the new article and close the second popup, but not the first popup.
> 
> _Same with opening first reference popup, then second article preview popup._

This should be fixed now. Please test carefully.

PR #103, opened 2020-01-24
Use XHR mozSystem to bypass CORS limitations

stephanebisson:


hueitan:
Can I test the difference between the changes and original fetch?

stephanebisson:
> Can I test the difference between the changes and original fetch?

In the browser the behavior should be identical since the mozSystem flag only works in the device.

As far as I can tell it helps follow the redirect but if you don't see the "no-extract" responses from the summary API then it would be hard to test. 

Other people have confirmed the issue but it is still mysterious.

hueitan:
surprisingly, I just got the `no-extract` result and it works after this change, and together with the previous canonical title fixed, then things work well now.

PR #104, opened 2020-01-24
Get language links from pcs

stephanebisson:
It is probably faster than MediaWiki API


stephanebisson:
It turns out there is more to download with this endpoint and we would have to zip it with units/allLanguages to get the languages display name. Not sure it's worth it.

hueitan:
ya, I also found the `metadata` lack of the language display name.

PR #105, opened 2020-01-24
Handle error when loading article

stephanebisson:
Show an error page if loading an article fail.

One way to test this is to use the search page to list articles, then turn off your wifi before trying to load one.

https://phabricator.wikimedia.org/T236897


stephanebisson:
> Do we support `Refresh` in the center key when having error?

Yeah, I'll look into adding it tomorrow.

stephanebisson:
Added the "refresh" option

src/components/Article.js
```diff
@@ -87,12 +88,16 @@ const ArticleSection = ({
 const ArticleInner = ({ lang, articleTitle, initialSubTitle }) => {
   const i18n = useI18n()
   const containerRef = useRef()
-  const article = useArticle(lang, articleTitle)
+  const [article, loadArticle] = useArticle(lang, articleTitle)
 
-  if (!article) {
+  if (article.loading) {
```
hueitan: we can keep this form without the `loading` key, so the loading animation shows by default when fetching the event.

src/hooks/useArticle.js
```diff
@@ -3,10 +3,10 @@ import { useI18n } from 'hooks'
 import { getArticle, getSuggestedArticles } from 'api'
 
 export const useArticle = (lang, title) => {
-  const [article, setArticle] = useState()
+  const [article, setArticle] = useState({ loading: true })
```
hueitan: same here, so we don't need to pass `loading` key here

hueitan:
when the user has slow internet connection, do you think it's better to show loading animation when clicking the refresh key? otherwise, user doesn't see any status when clicking the refresh key.

src/components/Article.js
stephanebisson: Done

PR #106, opened 2020-01-24
T241171 search offline

stephanebisson:
https://phabricator.wikimedia.org/T241171


hueitan:
oh wait, the code auto refresh the search query when it got online, so we don’t need the Refresh button, right?

stephanebisson:
I don't think we need "refresh" on search but maybe on the article page to retry fetching the article.

stephanebisson:
I don't think @jpita will be available this week so we can merge the PRs and let him test them on master when he has time.

PR #108, opened 2020-01-27
Remove unused goto.quickfacts()

stephanebisson:


hueitan:
We need to also remove Quick Facts item from the menu list when it doesn't exist.

Do you want to make the change here or another pr?

stephanebisson:
> We need to also remove Quick Facts item from the menu list when it doesn't exist.
> 
> Do you want to make the change here or another pr?

I'll make another PR for that in a few minutes.

PR #111, opened 2020-01-27
T236302 image gallery

hueitan:
Phabricator, https://phabricator.wikimedia.org/T236302

Things still in discussion or may need discussion - loading experience, image article view in block or inline

this version doesn't have Ellipsis at the end of the about description line, will try and see how we restrict the number of line to be 3 and has ellipsis at the end, in the meanwhile, we can test this


stephanebisson:
While some aspects may still be under discussion I think the code looks good and it's a very exciting feature. I'll merge it to allow more people playing with it.

PR #114, opened 2020-01-28
Header page with small icons

stephanebisson:
https://phabricator.wikimedia.org/T234435


hueitan:
wow! amazing! we should merge this asap so team can play with it

I'm merging this first and update with pr when having the feedback from team.

---

the case `Specular_reflection` in size `4` and `6` having weird page break

also `Rainbow flag (LGBT)` in size `default`

Capture

hueitan:
another thing, previously the smaller content can stay only 1 page, but it's now forced to become 2 pages, better we can ignore the extra margin-top for 1 page content.

example : http://localhost:8080/#/article/en/Roberts%2C_Wisconsin/External%20links

+++

> the case `Specular_reflection` in size `4` and `6` having weird page break
> 
> also `Rainbow flag (LGBT)` in size `default`
> 
> Capture

PR #118, opened 2020-01-30
Fix cypress tests

stephanebisson:
The 'check footer' test is currently failing because it is
expecting different suggested articles for en/Cat.

Changed the test to assert that there is 3 suggested articles
in the footer but not what they are. Specific suggestions
do change over time for reasons outside the scope of this
application.


stephanebisson:
I will merge this in the interest of time. @jpita @hueitan you can still take a look a share your feedback. Thanks

PR #121, opened 2020-01-31
Header page positioning when text resizes

stephanebisson:
After the initial rendering of the article header page,
The .card component is added a margin-top to make sure
the .intro component fits perfectly at the bottom of
the page.

On text resize (make larger, 6), the .intro component
overflows to the next page and its height is reported
as bigger than it actually is. This causes the algorithm
to push the intro all the way up.

This commit resets the .card margin-top to 0 to make sure
the .intro is entirely visible.

### DIFF
--- a/src/components/Article.js
+++ b/src/components/Article.js
@@ -74,12 +74,16 @@ const ArticleSection = ({
       titleNode.classList.add('clamp')
     }
     if (imageUrl) {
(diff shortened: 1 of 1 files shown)


hueitan:
does this also fix the content only has 1 page?

stephanebisson:
> does this also fix the content only has 1 page?

No. I brought it up on the task so we can discuss the potential solutions.

PR #124, opened 2020-02-05
Add Clamp to the description in gallery & Add Gallery item in Menu

hueitan:
more than 3 lines.

Capture

### DIFF
--- a/src/components/Gallery.js
+++ b/src/components/Gallery.js
@@ -1,11 +1,12 @@
 import { h } from 'preact'
-import { useState, useLayoutEffect } from 'preact/hooks'
+import { useState, useRef, useLayoutEffect } from 'preact/hooks'
 import { useI18n, useSoftkey, usePopup } from 'hooks'
 
 const MAX_DESCRIPTION_HEIGHT = 44
 
 const AboutContainer = ({ author, description, license, filePage, close }) => {
   const i18n = useI18n()
+  const containerRef = useRef()
 
   useSoftkey('About', {
     right: i18n.i18n('softkey-close'),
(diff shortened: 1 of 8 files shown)


hueitan:
This pr includes adding Gallery item into menu.

src/components/Gallery.js
```diff
@@ -12,6 +14,14 @@ const AboutContainer = ({ author, description, license, filePage, close }) => {
     onKeyLeft: () => { window.open(filePage) }
   })
 
+  useLayoutEffect(() => {
+    const descriptionNode = document.querySelector('.description')
```
stephanebisson: Please don't use global selectors. Put a ref on the top div or the specific element you need.

style/gallery.less
```diff
@@ -62,8 +62,29 @@
 		overflow: hidden;
 
 		&.description {
-			max-height: 3.5em;
-			overflow: hidden;
+			&.clamp {
```
stephanebisson: Could you look into creating a LESS mixin instead of copying it over?

hueitan:
can you give me the article name of your screenshot example?

src/components/Gallery.js
hueitan: true, I think the above screenshot caused by this global reason. thanks for the catching.

hueitan:
we cannot confirm the ellipsis at the end of the text in 3rd line, we can only insert `...` in the end not the end of the text, see the screenshot

Capture

---

I'm thinking of another solution of having opacity in the bottom text between 3rd and 4th lines

will try and see the possible way to do it then comment on phab.

stephanebisson:
> we cannot confirm the ellipsis at the end of the text in 3rd line, we can only insert `...` in the end not the end of the text

It's true, we have the same problem with clamping on the article title. I think we can let it go into QA and product sign off.

src/components/Gallery.js
```diff
@@ -1,9 +1,12 @@
 import { h } from 'preact'
-import { useState } from 'preact/hooks'
+import { useState, useRef, useLayoutEffect } from 'preact/hooks'
 import { useI18n, useSoftkey, usePopup } from 'hooks'
 
+const MAX_DESCRIPTION_HEIGHT = 44
```
stephanebisson: How did you arrive at that number? It should be exactly 3 x line-height

hueitan: I didn't set the line-height for the text here, and 44 is the best match with `...`

or I can set the `line-height: 15px` and `max-height: 45px`

PR #125, opened 2020-02-05
Use normalizedtitle instead displaytitle to prevent HTML content in toc

hueitan:
This fixes the problem of unable selecting the section in toc, we don't want the display title to be used, same to the remaining section.

https://phabricator.wikimedia.org/T243921#5844220

example : http://localhost:8080/#/article/en/Back_to_Earth_(Cat_Stevens_album)

_this doesn't fix the duplicate section key_


jpita:
is the main issue going to be fixed on this PR or a different one?

hueitan:
different pr

PR #127, opened 2020-02-06
Check anchor instead of title and some refactoring

hueitan:
Phab. https://phabricator.wikimedia.org/T243921

@jpita check carefully whether it goes to the correct section when open toc & select item from toc


src/utils/canonicalizeTitle.js
```diff
@@ -0,0 +1,4 @@
+
+export const canonicalizeTitle = title => {
```
stephanebisson: This is basically the same thing as `utils/normalizeTitle.js`[1] I would suggest keeping only one and fixing it.

[1] `normalizeTitle` is not implemented well. It will only replace the first space with an underscore. It should do something like `title.replace(/\s+/g, '_')`

hueitan: so there are two things

normalizeTitle : convert all _ to ' '
canonicalizeTitle : convert all ' ' to _

stephanebisson: You're right

stephanebisson:
@jpita I will merge this now to avoid code conflicts. Please test it carefully when you have time.

PR #128, opened 2020-02-06
Add brand in search placeholder to increase awareness

stephanebisson:
https://phabricator.wikimedia.org/T236262#5855658

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -36,7 +36,7 @@
     "softkey-search": "Search",
     "search-language-placeholder": "Search languages",
     "softkey-menu": "Menu",
-    "search-placeholder": "Search",
+    "search-placeholder": "Search Wikipedia",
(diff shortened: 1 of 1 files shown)


i18n/en.json
```diff
@@ -36,7 +36,7 @@
     "softkey-search": "Search",
     "search-language-placeholder": "Search languages",
     "softkey-menu": "Menu",
-    "search-placeholder": "Search",
+    "search-placeholder": "Search Wikipedia",
```
hueitan: this key also being used in language page, do we want to apply this to all?

stephanebisson: There is `search-language-placeholder` above, shouldn't this one be used on the language page?

hueitan: yes, language page should use `search-language-placeholder`

PR #131, opened 2020-02-06
InukaPageView instrumentation

stephanebisson:
Heavily inspired by https://github.com/wikimedia/mediawiki-extensions-WikimediaEvents/blob/master/modules/ext.wikimediaEvents/InukaPageView.js

Schema: https://meta.wikimedia.org/w/index.php?title=Schema:InukaPageView&oldid=19739286
Phabricator Link: https://phabricator.wikimedia.org/T242358

Todo
- [x] Handle visibility change. Keep track of how long the page is hidden.
- [x] Keep track of how many sections are viewed
- [x] Confirm whether we need `beforeunload` event in the device << it doesn't fire window.beforeunload, so we use `pagehide`
- [x] `referring_domain` - `kaios-app` can be the value within app
- [x] Confirm the values chosen for `base url`, `webHost` and `wiki` with analytics people on phab
    - [x] `webHost`, either `en.wikipedia.org` or `en.m.wikipedia.org`
    - [x] `wiki`, [language]wiki `enwiki` `hiwiki`
- [x] <s>Adjust UA</s>, add `app_version` key
- [x] <s>(TBC) access_method `mobile` , https://phabricator.wikimedia.org/T246309</s>

https://phabricator.wikimedia.org/T242358


src/hooks/useTracking.js
```diff
@@ -63,8 +63,14 @@ export const useTracking = (
...
   sectionCount = 0,
-  openedSections = 0
+  openedSections = {}
 ) => {
+  const trackingRef = useRef()
+
+  useEffect(() => {
+    trackingRef.openedSectionCount = Object.keys(openedSections).length
+  }, [openedSections])
+
```
hueitan: Note :

We cannot simply use `openedSections` here, the reason is given here https://github.com/facebook/react/issues/14965

the suggestion is using `useRef` to handle things like mutable value
https://reactjs.org/docs/hooks-reference.html#useref

hueitan:
NOTE: 
I don't find a way to send beacon when the app exit, and we also need to make sure the `back/exit` button works the same in different devices.

jpita:
` @hueitan hueitan marked this pull request as ready for review 8 days ago` 

what's the status of this?
should I check it somehow?

src/hooks/useTracking.js
```diff
@@ -0,0 +1,131 @@
+import { useRef, useEffect } from 'preact/hooks'
+
+const SCHEMA_NAME = 'InukaPageView'
+const SCHEMA_REV = 19739286
+const USER_ID_KEY = 'INUKA-PV-U'
+const SESSION_ID_KEY = 'INUKA-PV-S'
+const ONE_HOUR = 3600 * 1000
+const BASE_URL = 'https://en.wikipedia.org/beacon/event'
```
stephanebisson: Should this URL vary based on current language?

stephanebisson:
> * [ ]  (TBC) access_method `mobile` , https://phabricator.wikimedia.org/T246309
This is for the VirtualPageView schema. Unrelated to this.

src/hooks/useTracking.js
hueitan: Event Logging uses `webHost` `wiki` to determine the article language, upon checking with team there is no need to change the base url.

hueitan:
The code is updated with the latest Schema now.

src/hooks/useTracking.js
```diff
@@ -103,6 +103,7 @@ export const useTracking = (
           session_id: getSessionId(),
           pageview_token: pageviewToken,
           client_type: 'kaios-app',
+          app_version: '1.0.0',
```
stephanebisson: We should probably read this from package.json

hueitan: Great! This is added here https://github.com/wikimedia/wikipedia-kaios/pull/131/commits/6be5f9a7dbcd634d3dd815123698da2d671a682f

jpita:
is this sending metrics to somewhere I can check?

PR #137, opened 2020-02-11
add test to check that the quick facts page

jpita:
add test to check that the quick facts page is opening from:
* the main article page
* the article menu


jpita:
@hueitan 139

PR #138, opened 2020-02-11
Make the figure selectable and show gallery

hueitan:
Phabricator Link : https://phabricator.wikimedia.org/T244841

### Problem Statement

Currently, the gallery is available, but user unable to select image within the article page.

### Solution

Let the figure (image) selectable and show the gallery with the selected image.

### Note

| Image Hover | Not Hover |
| --- | --- |
| Capture | Capture2
 |


src/components/Article.js
```diff
@@ -59,6 +59,9 @@ const ArticleSection = ({
     section: ({ text, anchor }) => {
       // @todo styling to be confirmed with design
       confirmDialog({ message: i18n.i18n('confirm-section', text), onSubmit: () => goToSubpage({ title: anchor }) })
+    },
+    gallery: ({ fileName }) => {
```
stephanebisson: In other similar link handlers, the key is named after what the target is rather than what will happen when you select it. Following the same style, the key here could be named `image`.

hueitan: true, and make sense, gonna change it!!!

hueitan:
Quick fact image is now selectable! Thanks for the reminder @stephanebisson

.stylelintrc.json
```diff
@@ -3,6 +3,7 @@
   "rules": {
     "selector-attribute-quotes": null,
     "no-descending-specificity": null,
-    "declaration-no-important": null
+    "declaration-no-important": null,
+    "selector-type-no-unknown": null
```
stephanebisson: Do you think there is a way to make it understand that 'figure-inline' exist instead of globally disabling this rule?

src/components/Article.js
```diff
@@ -156,8 +159,14 @@ const ArticleInner = ({ lang, articleTitle, initialAnchor }) => {
...
-  const showGallery = () => {
-    showGalleryPopup({ items: article.media })
+  const showGallery = fileName => {
+    let startIndex = 0
+
+    if (fileName) {
+      startIndex = article.media.findIndex(media => media.canonicalizedTitle === fileName)
+    }
+
+    showGalleryPopup({ items: article.media, startIndex })
```
stephanebisson: You could give the component the target filename and let it figure out the index so you would have to duplicate the findIndex here and in quick facts.

src/hooks/useArticleLinksNavigation.js
```diff
@@ -124,6 +126,13 @@ const makeLinkClickEvent = link => {
   if (link.hash) {
     return { type: 'section', text: link.textContent, anchor: link.getAttribute('href').slice(1) }
   }
+
+  if (link.tagName === 'FIGURE' || link.tagName === 'FIGURE-INLINE') {
+    const aElement = link.querySelector('a')
+    const href = aElement.getAttribute('href')
+    const fileName = href.slice(6)
```
stephanebisson: The `slice(6)` could use a comment explaining what's going on.

style/article.less
```diff
@@ -121,6 +121,11 @@
 						background-color: #eaecf0;
 						margin-top: -10px;
 					}
+
+					&[ data-selected='true' ] {
```
stephanebisson: Is it possible that we're starting to have those duplicated quite a bit? If so, do you think they can be extracted to variables or mixins?

hueitan: I think we can set the global mixin for styling when `data-selected` is true, can do some customization when needed (for example figure doesn't need border)

hueitan: I have made this as variables, there are too many variations between text, image, card, so I'm not using mixin here.

PR #140, opened 2020-02-12
App Icons in different pixels device

hueitan:
Phabricator Link : https://phabricator.wikimedia.org/T244374

### Problem Statement

Based on the requirement of Jio and Kai Store, we need to support different pixels of App Icon.

### Solution

Therefore, we provide `56` `84` `112` in circle app icons.

### Note

You should see the app icon on the device.


hueitan:
@medied feel free to merge this if there isn't any problem with the code, as I don't see we can test it at this moment.

PR #142, opened 2020-02-13
Display 'About' button in gallery dynamically

medied:
Phabricator Link : https://phabricator.wikimedia.org/T244462

### Problem Statement

When an image in an article gallery is missing meta data such as a `description`, clicking on the 'About' button fails to render properly and throws an error on console

### Solution

Following a recommendation from @hueitan, in this PR we check if the image has a `description` field and only render the button if so. We could also check for other values such as `caption` `author` etc, but the `description` is the only one being read within the `AboutContainer` in line 23:

```
const descriptionNode = containerRef.current.querySelector('.description')
```


medied:
Good call @hueitan, opted for a little helper method `containsNecessaryFields` to check the minimum number of fields is present

src/components/Gallery.js
```diff
@@ -22,7 +22,7 @@ const AboutContainer = ({ author, description, license, filePage, close }) => {
 
     const descriptionNode = containerRef.current.querySelector('.description')
 
-    if (descriptionNode.getBoundingClientRect().height > MAX_DESCRIPTION_HEIGHT) {
+    if (descriptionNode && descriptionNode.getBoundingClientRect().height > MAX_DESCRIPTION_HEIGHT) {
```
medied: This just to avoid the `TypeError: descriptionNode is null`

PR #143, opened 2020-02-17
:bug: Use anchor instead title for sub page

hueitan:
Phabricator Link : -

### Problem Statement

Selecting the section doesn't work in the article

### Solution

This is the bug fixed of https://github.com/wikimedia/wikipedia-kaios/pull/127, as we change to `anchor` instead of using the section `title`

### DIFF
--- a/src/components/Article.js
+++ b/src/components/Article.js
@@ -58,7 +58,7 @@ const ArticleSection = ({
     },
     section: ({ text, anchor }) => {
       // @todo styling to be confirmed with design
-      confirmDialog({ message: i18n.i18n('confirm-section', text), onSubmit: () => goToSubpage({ title: anchor }) })
+      confirmDialog({ message: i18n.i18n('confirm-section', text), onSubmit: () => goToSubpage({ anchor }) })
     }
   }
 


medied:
To make sure I'm not missing context:

- Running app in English with branch `T243921-Bug-Fixed`
- Testing with browser view (Firefox and Chrome), not a device
- Search for "Catalonia"
- Click on "Sections"
- If I click on main section titles (for example "History" or "Geography"), I see the expected section [1]
- If I click on a sub page (for example "Middle Ages" or "Modern Era"), the app jumps to a part of the article within the right section but doesn't land exactly on the selected sub page. It lands a bit off, so if I move or scroll a bit I can eventually see the subpage [2]

I'd expect the app to land at the right sub page part, but is it possible I'm seeing this behavior because I'm testing on browser instead of device? Anyone else seeing this behavior on browser?

---

[1]—

 Screen Shot 2020-02-17 at 13 36 25

[2]— 

Screen Shot 2020-02-17 at 13 36 57

hueitan:
> To make sure I'm not missing context:

Neither browser nor device has this problem from my side.

It's possible that you adjust the screen size after loading the page, try again and see whether you have the same result without adjusting the screen size.

Furthermore, try to develop the app in Firefox OS around version ~48, don't use the latest version due to the KaiOS environment. Chrome can be used but not recommend due to the same reason.

medied:
Ah thanks @hueitan, adjusting the screen size to a Jio phone standard ([240 x 320](https://www.digit.in/mobile-phones/jio-phone-price-64901.html)) corrects the sub page rendering for me.

PR #144, opened 2020-02-17
Change All Caps language key

hueitan:
Phabricator Link : https://phabricator.wikimedia.org/T243413

### Problem Statement

There is ALL CAPS language key in en.json that is not recommended.

### Solution

Follow the same pattern as other language keys

### Note

This key is being used in Article Footer as the title keyword.


hueitan:
@jpita ready to tested, very quick one

PR #145, opened 2020-02-17
Add search wikipedia option to article menu

medied:
Phabricator Link: https://phabricator.wikimedia.org/T244483

### Problem Statement

Include "Search Wikipedia" in article menu for users to be able to easily access Search when they are within an article

### Solution

From the perspective of the codebase, the Search view is effectively the "home page". Idea in this PR then is to include one more menu item in `ArticleMenu.js` that routes to home

### Note

One thing I noticed: when a user is reading an article and then clicks to a related second article from within the first article, the menu in that second article currently shows "Previous article" option first at the top, and the "Search Wikipedia" article second. Not sure if this is an issue? Otherwise the "Search Wikipedia" option always shows on top.


cypress/page-objects/article-page.js
```diff
@@ -41,4 +41,22 @@ export class ArticlePage {
     cy.rightArrow()
     cy.enter()
   }
+
+  goToLanguageMenuFromArticleLandingPage () {
```
hueitan: Do you use this somewhere?

cypress/integration/article-view-spec.js
```diff
@@ -19,12 +19,12 @@ describe('Article view', () => {
   })
 
   it('change article language', () => {
+    console.log('start of test')
```
hueitan: console log here

src/components/ArticleMenu.js
```diff
@@ -45,6 +51,7 @@ export const ArticleMenu = ({
   }, [])
 
   const items = [
+    { title: i18n.i18n('search-placeholder'), action: onSearchSelected },
     { title: i18n.i18n('article-action-sections'), action: onTocSelected },
     { title: i18n.i18n('menu-textsize'), action: onTextsizeSelected }
   ]
```
hueitan: Thing come in my mind for the refactoring.

The list is getting bigger, instead of push & unshift to the list, we can use the full list and filter it based on the condition, a lot easier for us to know the order of the item.

jpita: should this be fixed by @medied on this PR or is just a comment for future work?

hueitan: Future

medied:
Alright so currently just [waiting to hear from Angie](https://phabricator.wikimedia.org/T244483) to confirm either merge or update.

src/components/ArticleMenu.js
```diff
@@ -51,46 +51,48 @@ export const ArticleMenu = ({
...
-  if (hasInfobox) {
-    items.push({
+    {
+      title: i18n.i18n('search-placeholder'),
+      action: onSearchSelected,
+      enabled: true
+    },
+    {
+      title: i18n.i18n('menu-previous'),
+      description: articleHistory.hasPrev() ? articleHistory.getPrev().title : null,
```
medied: Using ternary operator here to avoid this `TypeError` when there’s no previous article:

```
TypeError: _utils__WEBPACK_IMPORTED_MODULE_4__.articleHistory.getPrev(...) is undefined
```

jpita:
> Alright so currently just [waiting to hear from Angie](https://phabricator.wikimedia.org/T244483) to confirm either merge or update.

please bring this up (if I forget :D ) during the daily.
in any case, we should default to the designs to avoid losing time with this back and forth, my bad.

jpita:
> > Alright so currently just [waiting to hear from Angie](https://phabricator.wikimedia.org/T244483) to confirm either merge or update.
> 
> please bring this up (if I forget :D ) during the daily.
> in any case, we should default to the designs to avoid losing time with this back and forth, my bad.

Angie has spoken, search always on top like on designs, please fix @medied

jpita:
@medied found a big issue:
when there's no previous article, the menu actions are wrong, the "article size" option opens the sections page

jpita:
> @medied found a big issue:
> when there's no previous article, the menu actions are wrong, the "article size" option opens the sections page

@hueitan can you spot the code that does this and find a way to avoid this issue in the future?

src/components/ArticleMenu.js
hueitan: You don’t need the ternary here, just 

> articleHistory.getPrev().title

hueitan:
@jpita it's fixed now, you will need to fix the test error, similar to that `selectOption` to select previous article.

src/components/ArticleMenu.js
```diff
@@ -14,7 +14,8 @@ export const ArticleMenu = ({
   const i18n = useI18n()
   const onKeyCenter = () => {
     const { index } = getCurrent()
-    const item = items[index]
+    const enabledItems = items.filter(item => item.enabled)
+    const item = enabledItems[index]
```
medied: Hmm thanks for correcting this @hueitan

medied:
Ah good catch @jpita and thanks @hueitan for taking a look! "Search Wikipedia" always on top now 👍

PR #147, opened 2020-02-19
Interchange LSK and RSK actions

hueitan:
Phabricator Link : https://phabricator.wikimedia.org/T245438

### Problem Statement

Current app has the opposite behaviour of left and right key action

### Solution

Follows the rule

```
LSK: Close actions
RSK: Menu and settings options
```


hueitan:
oh.. @jpita you might need to change the test spec globally ...

PR #148, opened 2020-02-19
Fix White Blink when booting app

hueitan:
Phabricator Link : -

### Problem Statement

The browser/app are currently showing either white blank w (w/o) empty list result message when booting the app. (assume you have onboard the app before and then exit the app)

### Solution

- [x] Don't show the empty list result message in the beginning
- [x] Show the background color in `<body>`


hueitan:
@jpita I assign you to check this as well as you have the device to test it.

PR #150, opened 2020-02-20
T244568 missing goto quickfacts

medied:
Phabricator Link : https://phabricator.wikimedia.org/T244568

### Problem Statement

Missing `goToArticleSubpage` functionality within QuickFacts component

### Solution

Opting for passing in `goToArticleSubpage` directly to QuickFacts component and abstract all of that logic to the parent component Article. Does this make sense? I'm able to replicate to expected behavior with the "Holly" case but not the "C" case ([phab link](https://phabricator.wikimedia.org/T244568) for more info).


hueitan:
ohh... I see more problems here, the below shows three examples of "goto" section

- `<a href="#Selected_species">see text</a>`

The textContent is not the name of the section, probably use the hash and normalize it to `Selected species` is the best one, although the anchor may not always be the same as the name of section

- `<a href="#Related_characters" title="C" data-selected="true">See below</a>`

This doesn't work in here because article preview (it has attribute title) is being triggered first, we should reorder the logic and check hash before the article preview

- `<a href="#Controversies" data-selected="true">Controversies</a>`

This we don't have the problem

medied:
> The textContent is not the name of the section, probably use the hash and normalize it to Selected species is the best one, although the anchor may not always be the same as the name of section

Ah I see what you mean here: having the `confirmDialog` message say 'Go to Section "Selected species"' instead of 'Go to Section "see text"' like below.

image

Your comment pointed me in the right direction. I switched the hash conditional and put it before title in `makeLinkClickEvent` and that fixed the issue, but would cause links that are meant for another article to fail. For example: I open "Stomach" in `en` > go to quick facts > click on "Pyloric sphincter"; and the `confirmDialog` comes up but selecting it throws `TypeError: article.toc.find(...) is undefined`.

Opting to leave title conditional as first check in `makeLinkClickEvent` but `and` it with an extra condition: `link.pathname !== '/'`. Logic there is that if `pathname === '/'`  then it’s a link to a section within the article, so let hash conditional handle that case. Does that make sense to you?

src/hooks/useArticleLinksNavigation.js
```diff
@@ -124,7 +124,8 @@ const makeLinkClickEvent = link => {
   }
 
   if (link.hash) {
-    return { type: 'section', text: link.textContent, anchor: link.getAttribute('href').slice(1) }
+    const normalizedText = link.hash.slice(1).replace('_', ' ')
```
hueitan:
```suggestion
    import { normalizeTitle } from 'utils' // import in the beginning
    
    // then
    const normalizedText = normalizeTitle( link.hash.slice(1) )
```

medied: Ah good call, wasn't aware `normalizeTitle` existed 👍

medied:
> the `confirmDialog` replaces the previous popup, we will need to update the `utils.confirm` file and use `stack: true`

If I change `utils.confirm` to use `stack: true` the behavior changes: the `confirmDialog`comes up as usual but when I select to confirm navigation the URI updates accordingly but the screen stays the same. 

So not sure what I'm missing. What do you mean by "replaces the previous popup"?

hueitan:
> If I change `utils.confirm` to use `stack: true` the behavior changes: the `confirmDialog`comes up as usual but when I select to confirm navigation the URI updates accordingly but the screen stays the same.

This is not expected, when confirm (click enter) through the dialog in this case, it should **close all the popup**, the screen should stay at article screen, not quick fact screen

Popup accepts `close` and `closeAll`, it closes one popup or all popups, it's already implemented in popup container, but not `confirm dialog`, you will need to add extra prop for that to make it choose close or closeAll.

Let me know if the explanation still not clear.

medied:
Ah I think I see what you mean now by "the `confirmDialog` replaces the previous popup": when `confirmDialog` is triggered inside QuickFacts, the confirm popup shouldn't replace the quick facts popup [2], rather it should stack on top of it [1].

So I pushed [the last commit](https://github.com/wikimedia/wikipedia-kaios/pull/150/commits/f2b4a7196390036dc3d28e3e35faf1e20a110578) for the purpose of communication in this thread. Check this out: if I pass in `closeAll` as an extra prop to `confirmDialog`, the app does close both popups and lands in the correct article section, but the softkeys get messed up and I can't navigate anymore [3]. Does anyone else see the same behavior? Will continue diving into it to see what's going on. 

---

[1]—
image


[2]—
image

[3]—
image

hueitan:
a quick view into the code, there is a bug - softkey stack didn't clear correctly when `closeAll()`, that said opening article preview in quickfact and load the new article doesn't clear the softkey stack as well

I don't see the problem of `close` 

we can talk about this later and require more time to look into the details.

medied:
As I continue debugging I think I'm narrowing down the issue to `useArticleLinksNavigation`: it gets invoked within `QuickFacts` and `ArticleSection` and makes use of `useSoftkeys` to handle `onKeyArrowRight` and `onKeyArrowLeft` cases. If I try resetting the `left` and `right` fields as show in [last commit](https://github.com/wikimedia/wikipedia-kaios/pull/150/commits/25e397ee9372a413aa260495011927752897c223), it seems it's in the right direction as I get the expected soft keys but I don't like this solution. 

> softkey stack didn't clear correctly when closeAll()

I don't like this solution because I keep going back to what you said earlier. It feels backwards to have to reset `left` and `right` instead of making sure the softkey stack clears correctly. Not to mention that navigating with the arrow key down still feels buggy.

medied:
After syncing with Huei and in the spirit of staying agile and organized, I'm reverting changes back to the working solution without the popups stacking up and decoupling the popup stack issue into a separate task.

jpita:
@hueitan why did you merge and move the task into sign off?

hueitan:
@jpita I thought you finished with the testing, sorry for that, please open again if there are more things to do here.

PR #156, opened 2020-02-25
Add language for Maithili, Bodo, Goan Konkani

hueitan:
Phabricator Link : -

### Problem Statement

There are some missing language in our language list 

### Note

Maithili : मैथिली
Bodo : बड़ो
Goan Konkani (Devanagari script) : गोवा कोंकणी

### DIFF
--- a/src/utils/languages.js
+++ b/src/utils/languages.js
@@ -1440,6 +1440,21 @@ const languages = [
     canonical_name: 'Herero',
     code: 'hz',
     name: 'Otsiherero'
+  },
+  {
+    canonical_name: 'Maithili',
+    code: 'mai',
+    name: '\u092E\u0948\u0925\u093F\u0932\u0940'
+  },
+  {
+    canonical_name: 'Bodo',
+    code: 'brx',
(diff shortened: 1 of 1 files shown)


hueitan:
@jpita can you confirm this with the language team, so to make sure we don't add the wrong name or display name to the language?

src/utils/languages.js
```diff
@@ -1440,6 +1440,21 @@ const languages = [
     canonical_name: 'Herero',
     code: 'hz',
     name: 'Otsiherero'
+  },
+  {
+    canonical_name: 'Maithili',
+    code: 'mai',
+    name: '\u092E\u0948\u0925\u093F\u0932\u0940'
```
Nikerabbit: Is there a specific need to use unicode escapes? Most JSON parsers can deal with Unicode.

hueitan: not really, the initial list is given in the Unicode escape form, therefore we are doing the same here.

Abijeet:
The language codes used for the languages look fine to me.

PR #157, opened 2020-02-25
{T245966} Adjust extra key to allow gallery in different languages

hueitan:
Phabricator Link : https://phabricator.wikimedia.org/T245966

### Problem Statement

The Gallery only works well in language `en` and sometimes doesn't show image in some items

### Solution

1. capture the filename in `normalized` form
1. thumbnail can be missing in api, use `srcset` as the substitution
1. in case the image is not found in the gallery, show the first item to prevent script error

### Note

Please test well in different language, @jpita thanks for finding the bug here.

### DIFF
--- a/cypress/integration/special-cases.js
+++ b/cypress/integration/special-cases.js
@@ -28,7 +28,7 @@ describe('special cases tests', () => {
     articleMenuPage.getMenuOption('Language').should('not.exist')
   })
 
-  it('gallery opens from a non-english article', () => {
+  it.skip('gallery opens from a non-english article', () => {
     cy.navigateToPageWithoutOnboarding('article/pl/Tupolew_Tu-154')
     articlePage.title().should('have.text', 'Tu-154')
     for (let index = 0; index < 15; index++) {
--- a/cypress/integration/special-cases.js
+++ b/cypress/integration/special-cases.js
@@ -27,4 +27,16 @@ describe('special cases tests', () => {
     cy.clickMenuButton()
     articleMenuPage.getMenuOption('Language').should('not.exist')
   })
+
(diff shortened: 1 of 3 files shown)


jpita:
this seems to be working now here and was not working on master http://127.0.0.1:8080/#/article/pl/Tupolew_Tu-154

hueitan:
@jpita do you mind adding the test spec in the same branch?

cypress/integration/special-cases.js
```diff
@@ -27,4 +27,16 @@ describe('special cases tests', () => {
     cy.clickMenuButton()
     articleMenuPage.getMenuOption('Language').should('not.exist')
   })
+
+  it('gallery opens from a non-english article', () => {
+    cy.navigateToPageWithoutOnboarding('article/pl/Tupolew_Tu-154')
+    articlePage.title().should('have.text', 'Tu-154')
+    for (let index = 0; index < 15; index++) {
```
jpita: @hueitan I haven't found a way to make this test robust.
I can't do conditional testing with [cypress](https://docs.cypress.io/guides/core-concepts/conditional-testing.html#Definition), I wanted to do something like:
```
if(element is on screen)
  enter()
else
  downArrow()
```
I tried using document.querySelector but doesn't work as well.
also, since the page flipping on electron/chrome is not working properly I don't know for sure the number of page flips needed.
I tried with `index < 15` and it works on local but not on the CI.

hueitan: currently navigating doesn't work as expected, it only works for firefox

I don't have a quick fix for now, I suggest you put a comment on this skipped test so we can come back later when having the fix.

PR #158, opened 2020-02-26
Test text size change

jpita:


jpita:
> * Change the size in preview and reference and back to article
> *

@hueitan can u check now?

PR #159, opened 2020-02-26
Request native language name in getLanglinks

medied:
Phabricator Link : https://phabricator.wikimedia.org/T245712

### Problem Statement

As described in the [Phabricator ticket](https://phabricator.wikimedia.org/T245712), what's desired is that languages are searched with the native name ("english") instead of the localized name ("inglês")

### Solution

Switching `llprop` field from `langname` to `autonym` in `getLanglinks` api

### Note

https://www.mediawiki.org/wiki/API:Langlinks

### DIFF
--- a/src/api/getLanglinks.js
+++ b/src/api/getLanglinks.js
@@ -6,14 +6,14 @@ export const getLanglinks = (lang, title) => {
     titles: title,
     prop: 'langlinks',
     lllimit: 500,
-    llprop: 'langname'
+    llprop: 'autonym'
   }
   const url = buildMwApiUrl(lang, params)
   return cachedFetch(url, response => {
     const { pages } = response.query
     return pages[0].langlinks.map(item => (
       {
-        title: item.langname,
+        title: item.autonym,
         lang: item.lang,
         description: item.title
       }
--- a/src/api/getLanglinks.js
+++ b/src/api/getLanglinks.js
@@ -6,14 +6,14 @@ export const getLanglinks = (lang, title) => {
     titles: title,
     prop: 'langlinks',
     lllimit: 500,
-    llprop: 'autonym'
+    llprop: 'langname|autonym'
   }
   const url = buildMwApiUrl(lang, params)
   return cachedFetch(url, response => {
     const { pages } = response.query
     return pages[0].langlinks.map(item => (
       {
-        title: item.autonym,
+        title: [item.langname, item.autonym],
         lang: item.lang,
         description: item.title
       }
--- a/src/api/getLanglinks.js
+++ b/src/api/getLanglinks.js
@@ -13,7 +13,8 @@ export const getLanglinks = (lang, title) => {
     const { pages } = response.query
     return pages[0].langlinks.map(item => (
       {
-        title: [item.langname, item.autonym],
+        title: item.langname,
+        autonym: item.autonym,
         lang: item.lang,
         description: item.title
       }
--- a/src/api/getLanglinks.js
+++ b/src/api/getLanglinks.js
@@ -13,8 +13,8 @@ export const getLanglinks = (lang, title) => {
     const { pages } = response.query
     return pages[0].langlinks.map(item => (
       {
-        title: item.langname,
-        autonym: item.autonym,
+        title: item.autonym,
+        langname: item.langname,
         lang: item.lang,
         description: item.title
       }
--- a/src/components/RadioListView.js
+++ b/src/components/RadioListView.js
@@ -6,21 +6,17 @@ export const RadioListView = ({ items = [], header, containerRef, empty }) => {
       { header && <div class='header'>{header}</div> }
       <div class='list' ref={containerRef}>
         {
-          items.length ? items.map(item => {
-            let title
-            Array.isArray(item.title) ? title = item.title[0] : title = item.title
-            return (
-              <div class='item' data-selectable data-title={title} data-selected-key={title} key={title}>
-                <div class='info'>
-                  <div class='title'>{title}</div>
-                  { item.description && <div class='description'>{item.description}</div> }
-                </div>
-                <div class='radio-container'>
-                  <div class={`radio ${item.isSelected ? 'selected' : ''}`} />
-                </div>
+          items.length ? items.map(item => (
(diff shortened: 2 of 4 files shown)


jpita:
this is not working well.
if I :
1-change the app's language to Portuguese 
2-go to `gato` article 
3-go to the article language selection
4-type `ingl` 
the list is empty, it should show the English language.

From my pov, the language search should always be in the language of the app, if I'm wrong please let me know how it should work.

medied:
Yea this is an interesting case: what is the best user experience? Have the language search display localized or native languages? Or both?

> From my pov, the language search should always be in the language of the app

This contradicts the ticket initial request, which was to have “English” instead of “Ingles” come up in the search. I think the question we should ask ourselves then is: should the language search always *and only* be in the language of th app? Or are we talking about having both “English” (native) and “Ingles” (localised) work in the search?

If it’s the latter case, @hueitan does the `LangLinks` api support requesting both `autonym` and `langname` at the same time? What I gather from [the docs](https://www.mediawiki.org/wiki/API:Langlinks) is that it's either one or the other. If I quickly try out some things (`llprop: 'autonym | langname’`, or having two `llprop` keys), I'm not getting anything useful. Am I missing something? Seems like having both localized and native show up in the language search would represent a bigger change to the code and tests. 

This is an interesting pickle we got in our hands, team.

hueitan:
`llprop: 'autonym|langname'` will work to get both props in the response

the best experience for me is searching and filtered by all the props `title` `langname` `autonym`

but better to ask product and design to decide what's the "best" of search the article in different language.

medied:
@hueitan is right! After taking a closer look we can have the search bar respond to both localized and native forms of the language, which as discussed earlier we believe would be a better user experience. 


from `en` to `pt` | from `pt` to `en` | from `pt` to `en`
----------------- | ----------------- | ----------------- 
 |  | 


Hopefully the above is intuitive enough. However, something else we could do is have both versions of the language name display

from `en` to `pt` | from `pt` to `en` | from `pt` to `en`
----------------- | ----------------- | ----------------- 
 |  | 

As we can see, with the counterpart name in parenthesis the screen looks a bit more crowded but would it be more useful for the user this way? Not sure if we have data to answer that.

The latest commit is without the counterpart in parenthesis, but it's a doable change if we decide we want to add the parenthesis. What do you think?

hueitan:
I vote for the above solution. extra parenthesis + working is too expensive here

Nikerabbit:
FYI: UniversalLanguageSelector provides language search API that works with any language: https://en.wikipedia.org/wiki/Special:ApiSandbox#action=languagesearch&format=json&search=ingl&formatversion=2

jpita:
@medied what's the status of this PR?
are we going to use Niklas suggestion?

hueitan:
> FYI: UniversalLanguageSelector provides language search API that works with any language: https://en.wikipedia.org/wiki/Special:ApiSandbox#action=languagesearch&format=json&search=ingl&formatversion=2

Hi @Nikerabbit Thanks for the api suggestion here, we are looking for the best way to use langlinks. We probably should use languagesearch for our language change feature.

@medied please use the above solution in your comment.

@jpita not yet.

medied:
Cool, well if want to go ahead with the solution from last comment (the one without extra name in parenthesis) then that's already live from my [last commit](https://github.com/wikimedia/wikipedia-kaios/pull/159/commits/e5d89fde7b086fcae4325f865ecf99a0d43e5911). If there are no more changes requested and it's approved by @hueitan, then @jpita this would be ready for testing. 

Thank you @Nikerabbit for the API suggestion!

src/components/RadioListView.js
```diff
@@ -6,17 +6,21 @@ export const RadioListView = ({ items = [], header, containerRef, empty }) => {
...
       <div class='list' ref={containerRef}>
         {
-          items.length ? items.map(item => (
-            <div class='item' data-selectable data-title={item.title} data-selected-key={item.title} key={item.title}>
-              <div class='info'>
-                <div class='title'>{item.title}</div>
-                { item.description && <div class='description'>{item.description}</div> }
+          items.length ? items.map(item => {
+            let title
+            Array.isArray(item.title) ? title = item.title[0] : title = item.title
```
hueitan: We need to make sure `(Radio)ListView` is the reusable component, I don't suggest we introduce array to the prop title

src/api/getLanglinks.js
```diff
@@ -6,14 +6,14 @@ export const getLanglinks = (lang, title) => {
...
   }
   const url = buildMwApiUrl(lang, params)
   return cachedFetch(url, response => {
     const { pages } = response.query
     return pages[0].langlinks.map(item => (
       {
-        title: item.langname,
+        title: [item.langname, item.autonym],
         lang: item.lang,
         description: item.title
```
hueitan: I would suggest renaming all the key to `langname` `autonym` or other sensible names, then parse it to ListView in the form of `{ title, lang, description}` (this is used by ListView Component)

src/components/RadioListView.js
medied: Yea the reason I included the `Array.isArray()` check was to make sure the component remained reusable, otherwise it was failing tests. Your feedback was great though, you made me realize I can simplify logic to actually leave `RadioListView` untouched

jpita:
does this seem correct to you?
the language of the app is `spanish`.
the language of the article is `hindi`.
when I search for `espa` it shows the correct title of the article in spanish but the language `name` is in hindi, quite confusing (unless that's the wanted behaviour).

image

hueitan:
> does this seem correct to you?
> the language of the app is `spanish`.
> the language of the article is `hindi`.
> when I search for `espa` it shows the correct title of the article in spanish but the language `name` is in hindi, quite confusing (unless that's the wanted behaviour).

is should show `español` right? just checking on mobile web and yes. 

@medied so get autonym to be title, exchange it with langname should work then.

I think then we are good to go 😄

medied:
@jpita @hueitan good call, switched now

jpita:
this seems ok now.
should we create a task to use Niklas suggestion in the future so we can search in any language?

PR #161, opened 2020-02-27
Update viewport get method

hueitan:
Phabricator Link : -

### Problem Statement

we use `window.screen` to calculate the viewport size, it works well but fail in cypress ui test.

### Solution

Capture the article viewport

### Note

This is intended to fix for cypress test


hueitan:
oh no, currently our navigation doesn't work in chrome, will reopen this when works on the ui test on chrome.

PR #162, opened 2020-03-03
Set footer header according to article language

medied:
Phabricator Link: https://phabricator.wikimedia.org/T244524

### Problem Statement

"Suggested articles" label (footer header) should be displayed in the same language as the article language

### Solution (_outdated, see thread below for more info_)

The idea I want to present with this PR is using `loadMessages` to manually set the footer header according to article language. Some thoughts regarding this approach:

* If we end up removing `Language` from app menu as per [T244401](https://phabricator.wikimedia.org/T244401), my understanding is that shouldn't affect the setting of the footer label using the approach in this PR
* This approach takes care of setting footer label in article language for both the sections `tocWithFooter` and in the actual `ArticleFooter`
* If this approach is valid enough, should we get rid of the `toc-footer` keys in the `i18n` files?

Looking forward to feedback


src/hooks/useArticle.js
```diff
@@ -1,28 +1,27 @@
...
 
   const loadArticle = () => {
     setArticle(null)
-    Promise.all([getArticle(lang, title), getArticleMedia(lang, title), getSuggestedArticles(lang, title)])
-      .then(([article, media, suggestedArticles]) => {
+    Promise.all([getArticle(lang, title), getArticleMedia(lang, title), getSuggestedArticles(lang, title), loadMessages(lang)])
+      .then(([article, media, suggestedArticles, messages]) => {
         const { sections, toc } = article
-        const footerTitle = i18n.i18n('toc-footer')
+        const footerTitle = messages[lang] && messages[lang]['suggested-articles'] ? messages[lang]['suggested-articles'] : messages.en['suggested-articles']
```
stephanebisson: We shouldn't do raw access to the i18n files data. That's what [banana-i18n](https://github.com/wikimedia/banana-i18n) is for. It has a bunch of features including proper fallback chains. Not everything falls back to english immediately. 

I would suggest changing the way we load messages to just load them all on startup (assuming it doesn't introduce a performance hit) and changing the locale of the instance temporarily to resolve the message. Note that we would have to be careful not to leave the instance with the wrong locale after this point.

```
i18n.setLocale(article.contentLang)
const footerTitle = i18n.i18n('toc-footer')
// restore i18n locale
```

Alternatively, we could request a feature in banana-i18n to specify the locale for a specific `i18n()` call. That's a very common scenario in MediaWiki where we always have to think if a specific message is best displayed in user or content language.

hueitan: I'm thinking of is that possible we maintain two i18n object, one for app language, one for article language?

stephanebisson: You would have to create a new one every time you change the app or article language. I would suggest having only one instance with all messages in it. We already need to do it for the fallback chains to work correctly.

We have a related ticket to load messages faster: https://phabricator.wikimedia.org/T236898

hueitan: by doing that assuming we have one object contains all the messages `{ en: {}, de: {}, hi: {}, fr: {} ... }` we are able to load all `banana.load( <all messages>, <locale> )`

I don't see a way to load two different locales using single banana `banana.i18n()`, it doesn't
allow settings locale for each call of `i18n()`, `banana.setLocal()` can be used but we don't want to keep switching before we use for content or app language.

are there some other features I have missed in banana?

> Alternatively, we could request a feature in banana-i18n to specify the locale for a specific i18n() call. That's a very common scenario in MediaWiki where we always have to think if a specific message is best displayed in user or content language.

+1 for this as this will be very useful

medied: Great feedback all around, thank you guys. Just pushed a couple of commits that follow Stephane's suggestion. Few thoughts: 

* Using `setLocale` seems indeed cleaner inside `useArticle`
* Fallback chains seems to be working properly, for example:
    * bcl.json exists but doesn’t have `toc-footer` key
        * if I change cat article language to Central Bicol/Bikol, then label falls back to default  “Suggested articles” as expected
    * af.json does not exist
        * if I change cat article language to Afrikaans, then label falls back to “Suggested articles” as expected
* But am I missing something else regarding fallback chains?
* The performance seems to be _slightly_ hit on initial load but not in a deterrent way. What do you think? We can find a way to measure precisely if we want to be accurate

src/api/language.js
```diff
@@ -24,6 +30,21 @@ const loadMessages = (lang) => {
     .then((messages) => Object.assign.apply({}, messages))
 }
 
+const loadAllLanguagesMessages = () => {
```
medied: We could get rid off of this `loadAllLanguagesMessages` altogether, not sure if it's a handy helper to keep around

src/api/language.js
```diff
@@ -1,3 +1,9 @@
+import { allLanguages } from 'utils'
+import en from '../../i18n/en.json'
```
medied: Relatives paths like these okay in the codebase? Only saw one being used here:

https://github.com/wikimedia/wikipedia-kaios/blob/master/src/index.js#L7

medied:
As discussed in [PR#166](https://github.com/wikimedia/wikipedia-kaios/pull/166), introducing the [banana-i18n-loader](https://github.com/santhoshtr/banana-i18n-loader) to only load a set of languages instead of all

src/api/language.js
stephanebisson: It works but it's pretty ugly. You can setup an [alias](https://github.com/wikimedia/wikipedia-kaios/blob/master/webpack.config.js#L51) to make it nicer.

medied:
As suggested in [Phabricator ticket](https://phabricator.wikimedia.org/T244524), also setting extra article labels according to article language. Summed up they are:

* Footer labels: title, content license, view in broweser
* Article actions: Sections, Quick Facts, Gallery, Language

hueitan:
I'm okay with the Article actions.

I think it will be even nicer if we change the **footer section** to use content language, I'm looking the mobile web page as the reference, [cat in france](https://fr.m.wikipedia.org/wiki/Chat), the entire footer section shows the language in France, not only the "suggested articles" label

medied:
Gotcha, sounds good to me. Should I go ahead and `git revert` the commit b166cc9951d2359a6761eaeff275ac47df8e9f8a then?

> ...if we change the footer section to use content language

And then the extra labels in footer section is addressed with commit a073d93b977c36a67d0a9beb8c1d9df63fb5c0b5

stephanebisson:
> Gotcha, sounds good to me. Should I go ahead and `git revert` the commit [b166cc9](https://github.com/wikimedia/wikipedia-kaios/commit/b166cc9951d2359a6761eaeff275ac47df8e9f8a) then?

Probably

> 
> > ...if we change the footer section to use content language
> 
> And then the extra labels in footer section is addressed with commit [a073d93](https://github.com/wikimedia/wikipedia-kaios/commit/a073d93b977c36a67d0a9beb8c1d9df63fb5c0b5)

To be confirmed. Let's bring the conversation back to phab so all the right people can take part. It will be useful to document what the other platforms are doing with this.

In the meantime, we could bring this PR to its original purpose and wrap it up.

medied:
Great, I like the idea of bringing this PR back to its original purpose and wrapping it up. As suggested in Phabricator, now there exists separate tasks to evaluate the extra labels: [T247443](https://phabricator.wikimedia.org/T247443) and [T247444](https://phabricator.wikimedia.org/T247444). 

If we decide to go through with both or either of, then T247443 would more or less map to commit b166cc9951d2359a6761eaeff275ac47df8e9f8a and likewise T247444 with commit a073d93b977c36a67d0a9beb8c1d9df63fb5c0b5. The idea then is to keep this PR focused only on the "Suggested article" label. 

What do you think? Let me know if you'd like me to `git revert` commits b166cc9951d2359a6761eaeff275ac47df8e9f8a and a073d93b977c36a67d0a9beb8c1d9df63fb5c0b5

stephanebisson:
> [...] Let me know if you'd like me to `git revert` commits [b166cc9](https://github.com/wikimedia/wikipedia-kaios/commit/b166cc9951d2359a6761eaeff275ac47df8e9f8a) and [a073d93](https://github.com/wikimedia/wikipedia-kaios/commit/a073d93b977c36a67d0a9beb8c1d9df63fb5c0b5)

Feel free to organize branches and PRs in a way that is practical for you... As long as we have a simple version ("suggested articles" only) to review and merge.

medied:
Okay great, I simplified this PR back to its original purpose ("suggested articles" only) and it's ready for review

src/api/language.js
hueitan: should we remove this? I don't see how we can use this now and future.

hueitan:
When I tried to change the language to "Japanese", then refresh the page, the app(ui) language changes back to `en` although the localStorage still store `ja`, it seems like the load message doesn't work correctly.

Please correct me if I'm wrong, if we already have the banana-i18n-loader for all the i18n languages, do we still need to use `loadAllSupportedLanguage`?

src/api/language.js
medied: Yea I think we should, will push a commit for this

medied:
> When I tried to change the language to "Japanese", then refresh the page, the app(ui) language changes back to en although the localStorage still store ja, it seems like the load message doesn't work correctly.

Good catch Huei, this is unexpected behavior to me. Digging into this a bit more, what I'm seeing is that the language change on refresh only happens with languages that `loadSupportedLanguageMessages` does not provide (that is, languages not loaded with banana-i18-loader). If I change the language to any of the ones that `loadSupportedLanguageMessages` provides (currently `{ en, fr, es, hi, pt }`), then the language doesn't change on refresh. Need to dig into this a bit more, thoughts welcomed.

How does a page refresh translate to using the app on the device? Wondering out loud

> Please correct me if I'm wrong, if we already have the banana-i18n-loader for all the i18n languages, do we still need to use loadSupportedLanguageMessages?

Good question too, we should clarify. What I gathered from the [banana-i18n-loader docs](https://github.com/santhoshtr/banana-i18n-loader) is that it is a webpack loader for banana-i18n _message files_, but that we still need to `banana.load` those messages somewhere. Then `loadSupportedLanguageMessages` is simply providing the messages object in index.js, which is where the `banana.load` was originally occurring. Does that make sense?

jpita:
> How does a page refresh translate to using the app on the device? Wondering out loud

I hope in no way, otherwise there's a whole new behaviour for me to think about testing 😄

hueitan:
sorry for the confusing `refresh` term, in the device, you can exit the entire app, then start the app again

src/api/language.js
```diff
@@ -24,6 +29,11 @@ const loadMessages = (lang) => {
     .then((messages) => Object.assign.apply({}, messages))
 }
 
+const loadSupportedLanguageMessages = () => {
+  return { en, fr, es, hi, pt }
```
stephanebisson: From the latest conversation about language support, it sounds like we want to allow reading in all languages even if we know some of them will look broken. It means we have to load all `i18n/*.json` into the banana instance.

hueitan: as far as I understood, or maybe wrong, that's what `banana-i18n-loader` can do, right?

medied: Okay yea, loading all `i18n/*.json` should help with the refresh issue. And then yes, question is should still load all of them using `banana-i18n-loader` or using `fetch`. 

With `banana-i18n-loader` we would have a long series of imports in language.js, with fetch we could have something similar to commit e6f49b183f1fec7c87f132c8cd80e0a45975be20. Is there a better way?

src/api/language.js
```diff
@@ -24,6 +25,20 @@ const loadMessages = (lang) => {
     .then((messages) => Object.assign.apply({}, messages))
 }
 
+const loadAllLanguagesMessages = () => {
+  const messages = {}
+  allLanguages.forEach(language => {
+    try {
+      messages[language.lang] = require(`../../i18n/${language.lang}.json`)
```
stephanebisson: Can you simplify the path since you have the webpack alias?

medied:
Alright latest commit aims to load all languages using `require`. We need to make sure translation files are included in webpack bundle

src/api/language.js
```diff
@@ -24,6 +25,20 @@ const loadMessages = (lang) => {
...
+      // Translation not available, discard
+    }
+  })
+
+  return messages
+}
+
 export {
-  loadMessages
+  loadMessages,
```
stephanebisson: Is `loadMessages` still useful?

With all translations loaded from the start, the language selection component shouldn't have to explicitly anything anymore. Is there something that could be cleaned up there?

src/index.js
```diff
@@ -1,14 +1,15 @@
...
 import '../style/style.less'
 
 const lang = getAppLanguage() || navigator.language.substr(0, 2)
 const banana = new Banana(lang)
-loadMessages(lang).then((messages) => {
-  banana.load(messages)
-  render(<App i18n={banana} />, document.querySelector('.root'))
-})
+
+const allMessages = loadAllLanguagesMessages()
```
stephanebisson: Very minor: the const could be inlined here.

src/api/language.js
hueitan: yeah, this can be removed. 

so we already load all the language, we just need to `i18n.setLocale()` to set the new locale, and no more `loadMessages` needed.

hueitan: after this has been removed, we can remove `load` method then

src/api/language.js
```diff
@@ -24,6 +25,20 @@ const loadMessages = (lang) => {
...
+  loadAllLanguagesMessages
 }
```
hueitan: Once we have `loadAllLanguagesMessages` in this file, we are safe to move this under `utils`, as this becomes a method not related to `api`

src/components/ArticleFooter.js
```diff
@@ -2,14 +2,13 @@ import { h } from 'preact'
 import { useI18n } from 'hooks'
 import { buildWpMobileWebUrl, canonicalizeTitle } from 'utils'
 
-export const ArticleFooter = ({ lang, title, items = [] }) => {
+export const ArticleFooter = ({ lang, title, items = [], headerTitle }) => {
```
hueitan: the `lang` here already present as `contentLang`, so we can do the similar things in `useArticle` to have the interface language in `i18n.locale`, then use `setlocale( lang )` to get the header title here.

src/api/language.js
medied: The following is not working for me:

```
messages[language.lang] = require(`i18n/${language.lang}.json`)
```

Seems to be an open issue others are seeing: https://github.com/tleunen/babel-plugin-module-resolver/issues/291

Do you see the same behavior? Note this was working with `import` like we did with commit 3d2cb9e33f8747bb5d37d66c6580011a35bc3c03

If you're seeing the same behavior, I can go ahead and [add a comment](https://github.com/tleunen/babel-plugin-module-resolver/issues/291) to signal urgency

src/components/ArticleFooter.js
medied: Not a bad idea, this simplifies `Article.js` then

src/api/language.js
medied: Sounds good, how about including it directly under`src/utils/languages.js`: 00e08c4b2c1c79c953af456d787609344a443491

src/hooks/useArticle.js
```diff
@@ -12,17 +12,23 @@ export const useArticle = (lang, title) => {
     Promise.all([getArticle(lang, title), getArticleMedia(lang, title), getSuggestedArticles(lang, title)])
       .then(([article, media, suggestedArticles]) => {
         const { sections, toc } = article
+        const i18nLocale = i18n.locale
```
stephanebisson: Is this still needed here?

src/components/Language.js
```diff
@@ -20,13 +19,9 @@ export const Language = () => {
...
 
-      // load new language json file
-      loadMessages(item.lang).then((messages) => {
-        i18n.setLocale(item.lang)
-        i18n.load(messages)
-        setLang(i18n.locale)
-        setAppLanguage(item.lang)
-      })
+      i18n.setLocale(item.lang)
+      setLang(i18n.locale)
```
stephanebisson: I don't quite get why this line and the next and accessing lang in different ways.

src/components/ArticleFooter.js
```diff
@@ -4,12 +4,17 @@ import { buildWpMobileWebUrl, canonicalizeTitle } from 'utils'
 
 export const ArticleFooter = ({ lang, title, items = [] }) => {
   const i18n = useI18n()
-  const header = i18n.i18n('suggested-articles')
+  const i18nLocale = i18n.locale
+
+  i18n.setLocale(lang)
```
stephanebisson: Maybe put those 2 lines in a try block and the restoring of current language in finally?

src/hooks/useArticle.js
medied: Even though it seems a bit redundant to set and reset `locale` here after commit b0261c3f403a003c96e1170c380da4997ec7b217, we still need to do the brief `locale` switch in `useArticle` if we want the "Suggested articles" label to show in the article language in the section menu popup (becuase `sectionsWithFooter` and `tocWithFooter` are constructed here)

image

So it's a bit of a trade-off that comes from simplifying `Article.js` in commit b0261c3f403a003c96e1170c380da4997ec7b217. What do you think, is that commit/trade-off worth it?

hueitan: I think we can leave it here, in the long run, we should think of how to refactor to better use of i18n to separate app language and content language.

Currently, it is super unclear whether the lang is for app or content lang. The lang here is for content, but the other places are for app which confuse.

src/components/Language.js
hueitan: lang state here is only for `Language` component and allows softkey listen to lang dependency

medied: Hmm I don't quite get it either yet. It looks it's been this way since [the beginning of this component](https://github.com/wikimedia/wikipedia-kaios/commit/93f963caea15c89b1d3439824ea5e44d98022e1e#diff-ef8545b1621337d4615cbcf5bfb6f2dd). And what I'm seeing is that `item.lang` and `i18n.locale` are actually different when `onKeyCenter` is triggered

medied: Ohh sent my comment without seeing Huei's... makes sense, I see it in that dependency array of `useSoftkey`

jpita:
@medied is this "ok"?
image

PR #164, opened 2020-03-04
Minor tweak to use useRange

hueitan:
Phabricator Link: -

### Problem Statement

In Gallery, Onboarding and Article page, the code always checks whether it's on the given range.

### Solution

gives useRange a range, and don't repeat the condition checking code within the main code.

### Note

Minor tweak

Onboarding code doesn't change much, and this is ready for the upcoming onboarding change.


src/hooks/useArticlePagination.js
```diff
@@ -69,18 +68,12 @@ export const useArticlePagination = (
...
   }
 
   const showPrevSection = () => {
-    const prevSection = currentSection - 1
-    if (prevSection >= 0) {
-      setCurrentSection(prevSection)
-      setIsLastPage(true)
-    }
+    setCurrentSection(currentSection - 1)
+    setIsLastPage(true)
```
hueitan: oh wait, this need to be guard.

src/hooks/useRange.js
```diff
@@ -0,0 +1,17 @@
...
+  const setRangeValue = newValue => {
+    if (newValue < min) {
+      setValue(min)
+    } else if (newValue > max) {
+      setValue(max)
+    } else {
+      setValue(newValue)
+    }
+  }
+  return [value, setRangeValue]
```
stephanebisson: Would it make sense for `useRange` to expose `next()` and `previous()` functions?

hueitan: Good comment! It's totally make sense and that's the idea why another hook useRange here.

hueitan:
ready to be reviewed now.

PR #166, opened 2020-03-06
T244401 search language update

medied:
Phabricator Link: https://phabricator.wikimedia.org/T244401

### Problem Statement

Desired to switch the language change option out of the menu and to the home screen to make more accessible.

### Solution

Following Sudhanshu's design:

_latest version_:
image

_first version_:
image

### TODO:

- [x] Front end polishing
- [x] Make the interface language same as phone and default to English when the phone language is not supported in our app
- [x] Fix failing test in search-spec.js: `should show empty result found when search for "adf23cv1"`


cypress/integration/search-spec.js
```diff
@@ -46,10 +46,10 @@ describe('Article search', () => {
       .children().first().next().should('have.class', 'img')
   })
 
-  it('should show empty result found when search for "adf23cv1"', () => {
-    searchPage.search('adf23cv1')
-    searchPage.getEmptyContent().should('have.text', enJson['no-result-found'])
-  })
+  // it('should show empty result found when search for "adf23cv1"', () => {
```
medied: @jpita any idea why this test is failing? 

FYI: this is just a draft PR

jpita: no tests are failing in my local.
did u remove the test?

jpita: oh I see u commented the test.
next time just leave it failing in the CI until we find a cause, lost almost half an hour trying to understand which test u were talking about 😄 
so, ` searchPage.search('adf23cv1')` inserts that string in the search box, since now `1` goes to the language settings, it opens the settings...

I don't want to say "I told u so", but I did 😂

medied: Uhh that's a nasty bug 😄, like a UX bug even? Should probably make us reconsider. 

I'd have to look into it, maybe there's a way to only have "1" trigger the language change if user is not typing in the search bar? No idea how that would even work. But even then what about the case of a user trying to search "1" in search bar? 🤔

hueitan: anway, you can check `inputRef.current` whether is focus, similar to the center key example

```
center: current.type === 'DIV' ? i18n.i18n('centerkey-select') : ''
```

the design is changed and 1 is removed

cypress/integration/article-view-spec.js
```diff
@@ -56,10 +56,10 @@ describe('Article view', () => {
   it('check quick facts opens', () => {
     goToCatArticle()
     articlePage.selectOptionFromActionsMenu('quickfacts')
-    quickFactsPage.table().should('contains.text', 'Various types of domestic cat')
+    quickFactsPage.table().should('contains.text', 'Various types of the domestic cat')
```
medied: @jpita FYI I updated this test spec as someone changed the article content over the weekend and added a "the" to the quick facts:

https://en.wikipedia.org/w/index.php?title=Cat&action=history
https://en.wikipedia.org/w/index.php?title=Cat&oldid=944595654

medied:
By the way with the updated design using `Language` in the left softkey, we would be effectively taking care of [T245156](https://phabricator.wikimedia.org/T245156) with this same PR

hueitan:
the language page changes the app language, are we suppose it changes only the search languages?

cypress/integration/article-view-spec.js
hueitan: I'm putting this commit on master, otherwise keeps getting test spec error when developing others.

medied: Sounds good. I'm assuming you are talking about [a `git cherry-pick`](https://mattstauffer.com/blog/how-to-merge-only-specific-commits-from-a-pull-request/) or something like that?

hueitan: done and yes cherry-pick

medied:
> the language page changes the app language, are we suppose it changes only the languages for search result (content language)?

We should clarify this, yea. My understanding from [Angela's comment](https://phabricator.wikimedia.org/T244401#5857355) is that this change was only to touch the app language, not the article content language. 

Maybe it would help to summarize in words here what the status of the app would be if this PR is to be merged, it would be:

* There are two different ways to think about languages in the app: 1) the interface/app language and 2) the content/article language? 
* The interface/app language can be changed using the left softkey in the home search screen 
* The content/article language can be changed using the language option inside the article menu, so once user is already inside article reading. 

Does that make sense to everyone? Am I missing something?

medied:
Come to think of it, I wonder if it would be helpful to label article language change option inside article menu as "Article Language" instead of just "Language":

image

medied:
I'm marking this PR as ready for review now, with the caveat that the following requirement is still pending: 

>Make the interface language same as phone and _default to English when the phone language is not supported in our app_

Specifically the _italic_ part. Reason being that I think we need to consider the above with whatever is going to be the resolution of [PR#162](https://github.com/wikimedia/wikipedia-kaios/pull/162). Questions I currently have: 

* whether we end up using `loadAllLanguagesMessages` or `loadMessages`, we need to decide what's the correct order to load `new Banana(lang)`
* does [banana-i18n](https://github.com/wikimedia/banana-i18n) already handle a default to English if phone language is not supported?

I may be missing context, but looking forward to feedback on what you think it's the best way to proceed.

stephanebisson:
> * whether we end up using `loadAllLanguagesMessages` or `loadMessages`, we need to decide what's the correct order to load `new Banana(lang)`

I'm uneasy with loading ALL languages, even those we don't support, but I think we need to load all the languages we positively support. However, it should be done with webpack, not `fetch`. There's a [loader](https://github.com/santhoshtr/banana-i18n-loader) for that.

Could you explain the question about the "correct order"?

> * does [banana-i18n](https://github.com/wikimedia/banana-i18n) already handle a default to English if phone language is not supported?

Assuming English is loaded, yes.

hueitan:
> Make the interface language same as phone and default to English when the phone language is not supported in our app

https://github.com/wikimedia/wikipedia-kaios/blob/ebd2c340e55963527eab6f44755c08dc49a13764/src/index.js#L9

Here's the code checking the default lang of the app

medied:
Awesome thank you both.

>Could you explain the question about the "correct order"?

I think you two answered my question, I was making it more complicated in my head. Here's what I will do: 

* I will introduce the webpack loader to load languages we positively support 
* Then with `const lang = getAppLanguage() || navigator.language.substr(0, 2)` and making sure English is loaded, we can leave banana-i18n to handle the English default if phone language is not supported
* I will introduce these changes in the other branch however: `T244524-suggested-articles-in-content-language` (from PR#162). We can leave this (PR#166) branch as is for now, then figure out what to merge first when the time comes. Sound good?

Can we come to a consensus of what are the languages we positively support though? Should that be languages that have the `toc-footer`/`suggested-articles` label already?

medied:
Would having the "Search Wikipedia" placeholder text in the center help? 🤔

Screen Shot 2020-03-10 at 15 52 54

hueitan:
it's better to ping design for the landscape design

the styling for the gap between header and list move again (this time is smaller), we have the previous pr https://github.com/wikimedia/wikipedia-kaios/pull/165, can you see why it changes the sytling for the gap?

hueitan:
@jpita ready to be reviewed now.

medied:
@jpita @hueitan but let's not merged until the conversation in [T244401](https://phabricator.wikimedia.org/T244401) is settled.

jpita:
moving the phab task to "Needs Design"

medied:
After the team revised [T244401](https://phabricator.wikimedia.org/T244401), it was decided that we go with [T248080](https://phabricator.wikimedia.org/T248080#5988590) as a shorter term solution. 

Therefore, I propose we close this PR in favor of https://github.com/wikimedia/wikipedia-kaios/pull/188

PR #167, opened 2020-03-06
Remove font family and variable

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T245746

### Problem Statement

Currently, the app download the font family "Lato" from the google apis, and it against WMF user privacy policy

### Solution

Upon checking on the default font styling in several devices, the font works well without assigning any font family, therefore the decision is to use the default font in each device.

### Note

If there is a source pass the policy, that's also not a bad idea to use.

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,7 +8,6 @@
     />
     <meta name="theme-color" content="rgb(51, 102, 204)" />
     <title>Wikipedia</title>
-    <link href="https://fonts.googleapis.com/css?family=Lato&display=swap" rel="stylesheet">
   </head>
 
   <body>
--- a/style/article.less
+++ b/style/article.less
@@ -34,7 +34,6 @@
 			opacity: 0.85;
 
 			.title {
-				font-family: @titleFont;
 				font-size: 22px;
 				line-height: 28px;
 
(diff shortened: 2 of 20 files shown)


medied:
>it against WMF user privacy policy

And is Sudhanshu or someone from the design team aware of this fact? In the latest [Zeplin designs](https://app.zeplin.io/project/58dc46f4a83d1e477dd83859/screen/5e6297f199f89b1879326f1d) Sudhanshu was using a font family I believe. Can you point us to a link for the user privacy policy?

stephanebisson:
> > it against WMF user privacy policy
> 
> And is Sudhanshu or someone from the design team aware of this fact? In the latest [Zeplin designs](https://app.zeplin.io/project/58dc46f4a83d1e477dd83859/screen/5e6297f199f89b1879326f1d) Sudhanshu was using a font family I believe. Can you point us to a link for the user privacy policy?

It's not using the font that's problematic but pulling it from google at runtime. It allows them to track our users.

hueitan:
> someone from the design team aware of this fact

yes, we do check on the default font for different device and we don't see any problem yet.

medied:
Gotcha! Maybe I'm just used to seeing the app with Lato font and find it aesthetically better, but code-wise changes look good.

hueitan:
@jpita let us know if you see some screen behave weird without assigning font family

PR #169, opened 2020-03-09
Abortable XMLHttpRequest

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T244021

### Problem Statement

1. Typing in search bar trigger unnecessary fetch event when there is a new search term
1. Press random article multiple times has console error 
1. There isn't abort/cancel event in `cachedFetch`

### Solution

Implement `xhr.abort` in `cachedFetch`

### Note

currently, the random article uses `cachedFetch` with an extra key to make it non-cacheable fetch event,  we can make the fetch event more general to achieve cacheable and non-cacheable event but it requires more depth consideration of how we work with our cache mechanism and ajax (get) event.

Please test the app carefully with this update

### DIFF
--- a/src/utils/cachedFetch.js
+++ b/src/utils/cachedFetch.js
@@ -2,21 +2,42 @@
 // the last N requests to keep memory usage
 // under control.
 const requestCache = {}
-export const cachedFetch = (url, transformFn) => {
-  if (requestCache[url]) {
+const xhrList = {}
+
+export const cachedFetch = (url, transformFn, abortAllXhr = false, cache = true) => {
+  if (abortAllXhr) {
+    abortAllFetch()
+  }
+
+  if (cache && requestCache[url]) {
     return Promise.resolve(requestCache[url])
   }
 
   return new Promise((resolve, reject) => {
     const xhr = new XMLHttpRequest({ mozSystem: true })
+    const timestamp = (new Date()).valueOf()
+    xhrList[timestamp] = xhr
     xhr.responseType = 'json'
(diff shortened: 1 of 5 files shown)


hueitan:
☝️ @medied I just stole your commit 😄

src/utils/cachedFetch.js
```diff
@@ -2,21 +2,42 @@
...
+    abortAllFetch()
+  }
+
+  if (cache && requestCache[url]) {
     return Promise.resolve(requestCache[url])
   }
 
   return new Promise((resolve, reject) => {
     const xhr = new XMLHttpRequest({ mozSystem: true })
+    const timestamp = (new Date()).valueOf()
```
stephanebisson: `Date.now()` is equivalent

hueitan:
I can't reproduce the error log when hitting close before the article page in my device.

stephanebisson:
> I can't reproduce the error log when hitting close before the article page in my device.

You're right. The issue is still visible in the browser but not on the device.

However I think the fix is accidental: leaving the article page early to go back to search triggers a re-issue of the search query, which cancels all previous requests and prevents getArticle from resolving. If there was a way to leave the article page early and go somewhere else that doesn't trigger a request, the issue would still be there. The deterministic way to fix that is by ensuring that components that depend on data cancel their own requests when unmounting. 

I'm going to merge this in the interest of time to make sure we don't release with this issue but keep the phab task open for us to revisit if/when we have time.

PR #171, opened 2020-03-11
Add app indicator to user-agent

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T244547

### Problem Statement

The user-agent has to contain an app identifier in order to count requests to the page content service as page views.

### Solution

Add WikipediaApp/<version> at the beginning of the user agent for all API calls.

### DIFF
--- a/src/utils/cachedFetch.js
+++ b/src/utils/cachedFetch.js
@@ -19,6 +19,7 @@ export const cachedFetch = (url, transformFn, abortAllXhr = false, cache = true)
     xhrList[timestamp] = xhr
     xhr.responseType = 'json'
     xhr.open('GET', url)
+    xhr.setRequestHeader('User-Agent', navigator.userAgent + ' WikipediaApp/1.0')
     xhr.send()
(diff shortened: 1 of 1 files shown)


hueitan:
Test on the device, I can see the Header User-Agent being updated

```
Mozilla/5.0 (Mobile; Nokia_2720_Flip; rv:48.0) Gecko/48.0 Firefox/48.0 KAIOS/2.5.2 WikipediaApp/1.0
```

milimetric:
@stephanebisson & @hueitan: the commit message mentions and the code seems to confirm that WikipediaApp/ is at the beginning of the UA, but https://github.com/wikimedia/wikipedia-kaios/pull/171#issuecomment-597778529 shows it at the end.  Is it at the beginning (this would be better)?

stephanebisson:
@milimetric it was at the end when @hueitan tested it but I moved it to the beginning when I saw this comment: https://phabricator.wikimedia.org/T244547#5865738

PR #172, opened 2020-03-11
deploy task builds first

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T247381

### Problem Statement

When using the deploy task, one can forget to rebuild the app and push outdated code to the device. It leads to confusion and waste of time.

### Solution

Bundle the build task inside the deploy task.


medied:
@jpita @stephanebisson so now that there's no need to `npm run build`, should I go ahead and update README?

https://github.com/wikimedia/wikipedia-kaios/blob/master/README.md#build-the-app

stephanebisson:
> @jpita @stephanebisson so now that there's no need to `npm run build`, should I go ahead and update README?
> 
> https://github.com/wikimedia/wikipedia-kaios/blob/master/README.md#build-the-app

Good idea

medied:
Okay, I'll go ahead and do it with a PR just to keep things organized

PR #173, opened 2020-03-12
Add close event to Loading component

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T245240

### Problem Statement

User unable to exit from the loading state on article language page when there is an error during the api event

### Solution

make sure that the softkey is being called after the loading component, and adding the `onClose` event to loading component to overwrite the default behaviour

### Note

Interesting task that cleanup my mind how we use the softkey in the correct order


src/components/ArticleLanguage.js
```diff
@@ -9,12 +9,12 @@ export const ArticleLanguage = ({ lang, title, close }) => {
...
   const [articleLang, setArticleLang] = useState(lang)
   const [items, query, setQuery, numOfLanglink] = useSearchArticleLanguage(articleLang, title)
-  const [, setNavigation, getCurrent] = useNavigation('ArticleLanguage', containerRef, 'y')
 
   if (!items.length && !query) {
-    return <Loading message={i18n.i18n('article-language-loading-message')} />
+    return <Loading message={i18n.i18n('article-language-loading-message')} onClose={close} />
   }
 
+  const [, setNavigation, getCurrent] = useNavigation('ArticleLanguage', containerRef, 'y')
```
hueitan: **_Programatically solve the problem in reducer pop:_** 
we can move this back above of Loading component when we have this merged https://github.com/wikimedia/wikipedia-kaios/pull/174

or;

**_Ideally, in theory, follows the right order,_**
we can leave it here, so we have the correct order of softkey Loading first then ArticleLanguage, as `useNavigation` has softkey with origin `ArticleLanguage`

medied: Is there an advantage to moving it back above `Loading` component?

hueitan: Code looks prettier? Maybe?

medied: Yea leaving it as it is right now with `Loading` first makes sense to me

medied:
> Interesting task that cleanup my mind how we use the softkey in the correct order

Yea I think I'm still learning what exactly is the correct order

hueitan:
@jpita 
you need to press Left Softkey close to back to previous screen - article page

PR #175, opened 2020-03-13
Remove help and feedback from the settings

hueitan:
Phabricator Link: -

### Note

A very quick one removing it from the settings list


i18n/qqq.json
```diff
@@ -65,7 +65,6 @@
 	"settings-privacy": "Label of the Privacy policy item list. Shown in the Settings menu, redirect to the privacy policy browser page.",
 	"settings-term": "Label of Terms of use. Shown in the Settings menu, redirect to the Terms of use browser page.",
 	"settings-rate": "Label of the Rate item list. Shown in the Settings menu, used in rating the app.",
-	"settings-help-feedback": "Label of Help and feedback item list. Shown in the Settings menu, to find help and give feedback.",
```
stephanebisson: When removing message keys, you have to remove them from both qqq and en files. However, I think we should keep the keys in this case because it is already translated and I think we will include this feature at some point soon.

hueitan: Done and leave it as comment so we don't forget the key name

hueitan:
@jpita if no require test, we will merge this.

PR #176, opened 2020-03-13
Only add UA in prod

stephanebisson:

### Problem Statement

The UA header is explicitly added to XHR calls to enable counting those as page views. However, this header causes CORS error locally.

### Solution

With the webpack EnvironmentPlugin, we know whether the code is running in dev or prod mode so we can conditionally add the header.

### Note


medied:
Okay perfect, I like what you did with the second commit. Tested on Firefox as well as banana phone and app is working for me 👍

PR #179, opened 2020-03-17
changing cat to dog

jpita:
testing


cypress/integration/article-view-spec.js
```diff
@@ -42,7 +42,7 @@ describe('Article view', () => {
   })
 
   it('check footer changes language', () => {
-    goToCatArticle()
+    goToDogArticle()
```
jpita: i don't like dogs, put cats back

PR #181, opened 2020-03-18
Add AboutApp page

medied:
"About the app" page from the settings menu is introduced in this PR following [the design spec](https://app.zeplin.io/project/58dc46f4a83d1e477dd83859/screen/5e68a94ac447ac1348d0ff09). I tested it on device as well, please do the same to make sure it looks the way we need it to.

### Browser view

Portrait | Landscape
----------- | --------
  | 

### Note

Phabricator Link: https://phabricator.wikimedia.org/T236319


src/components/AboutApp.js
```diff
@@ -0,0 +1,34 @@
...
+  useSoftkey('AboutApp', {
+    right: i18n.i18n('softkey-read-more'),
+    onKeyRight: () => window.open('https://wikimediafoundation.org/'),
+    left: i18n.i18n('softkey-close'),
+    onKeyLeft: () => history.back()
+  }, [])
+
+  return (
+    <div class='aboutapp'>
+      <div class='header'>{i18n.i18n('gallery-about-header')}</div>
```
medied: Reusing `gallery-about-header` although the name may be slightly confusing here

src/components/AboutApp.js
```diff
@@ -0,0 +1,34 @@
+import { h } from 'preact'
+import { useI18n, useSoftkey } from 'hooks'
+
+export const AboutApp = () => {
+  const i18n = useI18n()
+  const APP_VERSION = JSON.stringify(require('../../package.json').version).slice(1, -1)
```
medied: I tried using the `DefinePlugin` global variable here (mirroring [this use of it](https://github.com/wikimedia/wikipedia-kaios/blob/master/src/hooks/useTracking.js#L107)), but it wasn't working for some reason.

Note I'm using a relative path within `require` because of the same issue [brought up here](https://github.com/wikimedia/wikipedia-kaios/pull/162#discussion_r392731983)

src/components/AboutApp.js
hueitan: I think it's better to have a different key here as they are different

src/components/AboutApp.js
hueitan: you can use 

```html
{/* eslint-disable-next-line no-undef */}
<p>{APP_VERSION}</p>
```

within the jsx line

i18n/en.json
```diff
@@ -71,5 +71,8 @@
     "textsize-decrease": "Press 4 (Decrease size)",
     "textsize-default": "Press 5 (Default)",
     "toc-footer": "Suggested articles",
-    "view-in-browser": "View article in browser"
+    "view-in-browser": "View article in browser",
+    "about-app-message": "Made by the Wikimedia Foundation with the help of volunteers like you.",
+    "softkey-read-more": "Read More"
+
```
hueitan: extra line

src/components/AboutApp.js
medied: Yea I suppose it doesn't cost us anything to add another key with the same text, just slightly more redundancy I guess. What about renaming `gallery-about-header` to something like just `about-header`?

hueitan: `about-header` looks good =)

src/components/AboutApp.js
stephanebisson: Right, you need to use the already-defined constant because package.json is not going to be available when the app runs on the phone.

src/components/AboutApp.js
stephanebisson: Renaming is fine in this case. TWN can handle that without losing all the translations.

src/components/AboutApp.js
medied: Perfect

src/components/AboutApp.js
```diff
@@ -23,6 +22,7 @@ export const AboutApp = () => {
           <img src='/images/wikipedia-wordmark-en.png' />
         </div>
         <div class='version'>
+          {/* eslint-disable-next-line no-undef */}
           <p>{APP_VERSION}</p>
```
stephanebisson: Now would be a good time to make that app version available through a utils and have that eslint suppression in one place only.

style/style.less
```diff
@@ -29,3 +29,4 @@
 @import './settings';
 @import './textsize';
 @import './toc';
+@import './aboutapp';
```
stephanebisson: Alphabetical order please ;)

src/utils/index.js
```diff
@@ -11,3 +11,4 @@ export * from './mwApi'
 export * from './normalizeTitle'
 export * from './onboarding'
 export * from './viewport'
+export * from './appVersion'
```
stephanebisson: Try to keep the index in alphabetical order please ;)

src/utils/appVersion.js
```diff
@@ -0,0 +1,4 @@
+export const appVersion = () => {
+  /* eslint-disable-next-line no-undef */
+  return APP_VERSION
```
stephanebisson: Other usages of APP_VERSION should now use the util.

hueitan:
It looks good, just the body padding-top 40px should be 30px according to the design, i think it's good to save some extra space for the translation.

| 40px | 30px |
| --- | --- |
| Screen Shot 2020-03-18 at 20 09 15 | Screen Shot 2020-03-18 at 20 09 11 |

medied:
Ah @hueitan yea good idea: will reduce padding a bit. It will look more centered on device this way, also

stephanebisson:
Approved 2 days ago.

jpita:
I keep seeing this now, can we fix this?
image

medied:
In which branch are you seeing this issue?

jpita:
> In which branch are you seeing this issue?

can't see it anymore, ignore my comment

PR #185, opened 2020-03-19
Fixed xhr error handling and error component background

stephanebisson:
### Problem Statement

Since moving from `fetch` to `XMLHttpRequest`, error handling has been partly broken since most error cases call the load callback instead of the error callback.

### Solution

Check the `status` and reject if it's not within [200-400[.

Also set the error component background to white.

### DIFF
--- a/src/utils/cachedFetch.js
+++ b/src/utils/cachedFetch.js
@@ -27,7 +27,8 @@ export const cachedFetch = (url, transformFn, abortAllXhr = false, cache = true)
     }
     xhr.send()
     xhr.addEventListener('load', () => {
-      if (xhr.statusText === 'OK') {
+      // Accept all of 2xx and 3xx
+      if (xhr.status >= 200 && xhr.status < 400) {
         const transformResponse = transformFn(xhr.response)
         resolve(transformResponse)
         if (cache) {
--- a/src/utils/cachedFetch.js
+++ b/src/utils/cachedFetch.js
@@ -27,10 +27,14 @@ export const cachedFetch = (url, transformFn, abortAllXhr = false, cache = true)
     }
     xhr.send()
     xhr.addEventListener('load', () => {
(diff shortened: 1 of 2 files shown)


src/utils/cachedFetch.js
```diff
@@ -27,10 +27,14 @@ export const cachedFetch = (url, transformFn, abortAllXhr = false, cache = true)
     }
     xhr.send()
     xhr.addEventListener('load', () => {
```
hueitan:
```suggestion
    xhr.addEventListener('load', event => {
```

src/utils/cachedFetch.js
```diff
@@ -27,10 +27,14 @@ export const cachedFetch = (url, transformFn, abortAllXhr = false, cache = true)
...
-      const transformResponse = transformFn(xhr.response)
-      resolve(transformResponse)
-      if (cache) {
-        requestCache[url] = transformResponse
+      if (xhr.statusText === 'OK') {
```
hueitan:
```suggestion
      if (event.target.statusText === 'OK') {
```

src/utils/cachedFetch.js
```diff
@@ -27,10 +27,14 @@ export const cachedFetch = (url, transformFn, abortAllXhr = false, cache = true)
...
+        const transformResponse = transformFn(xhr.response)
+        resolve(transformResponse)
+        if (cache) {
+          requestCache[url] = transformResponse
+        }
+      } else {
+        reject(new Error('XHR Error: ' + xhr.status))
```
hueitan:
```suggestion
        reject(new Error('XHR Error: ' + event.target.status))
```

src/utils/cachedFetch.js
stephanebisson: Does it change anything?

hueitan: nope, just guessing whether it helps.

PR #186, opened 2020-03-19
T247616 Show wikitable in article and expandable in new popup page

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T247616

### Problem Statement

The table is hidden within the entire article page.

### Solution

Show and allow enter for wiki table. (table has class name `wikitable` `multicol`

Not the elegant tech solution, but minimum changes to the html string and without creating extra object to handle

Modify the caption of each table and programmatically use class name `hidden-in-table` + css to hide unnecessary content

### Note

Final design still pending from Design.

| Table | Table in Article |
| --- | --- |
| Screen Shot 2020-03-19 at 21 30 38 | Screen Shot 2020-03-19 at 21 30 35 |


i18n/qqq.json
```diff
@@ -29,6 +29,7 @@
 	"header-textsize": "The header title of the text size page",
 	"menu-previous": "Label of the Previous article item. Shown in the Menu page.",
 	"menu-textsize": "Label of the Text size item list. Shown in the Menu menu.",
+	"more-information": "Label of the Text \"more information\". Used in Collapsed Table in Article page.",
```
stephanebisson: The first part is redundant since translator will see the English version on TWN. You can specify that selecting this element will open the table full screen in a new popup.

src/hooks/useArticle.js
```diff
@@ -42,3 +47,24 @@ export const useArticle = (lang, title) => {
 
   return [article, loadArticle]
 }
+
+const fixTableCaption = (htmlString, i18n) => {
+  const moreInformationText = i18n.i18n('more-information')
```
stephanebisson: Ideally, this line should be in the <Table> component since this is building the UI and everything else in this function should be in the transform function in `api/getArticle.js` so it is cached correctly.

src/hooks/useArticle.js
```diff
@@ -42,3 +47,24 @@ export const useArticle = (lang, title) => {
...
+  const hiddenClassName = 'hidden-in-table'
+  const parser = new DOMParser()
+  const node = parser.parseFromString(htmlString, 'text/html')
+  const tableNodes = node.querySelectorAll('table.wikitable, table.multicol')
+  for (const tableNode of tableNodes) {
+    const thContent = Array.from(tableNode.querySelectorAll('th')).map(th => th.textContent).join(', ')
+    const normalizedThContent = thContent.replace(/\[\d+]/g, '')
+    if (tableNode.caption && tableNode.caption.textContent) {
+      tableNode.caption.innerHTML = `<b class='${hiddenClassName}'>${moreInformationText}:</b><p class='${hiddenClassName}'>${normalizedThContent}</p><span>${tableNode.caption.textContent}</span>`
```
stephanebisson: This building of the UI definitely should be in <Table>

src/hooks/useArticle.js
hueitan: Initially, I also think it should be in `getArticle`, but then the same reason of having correct language translation for the footer is being built here.

I think it's fine we use i18n in `getArticle` so we can cache the result there.

hueitan: `more-information` is needed only in Article page, not Table.

hueitan: `more-information` is needed only in Article page, not Table.

hueitan: so, I have been trying to use `useI18n()` in `getArticle`, but got error and against the [rules of Only Call Hooks at the Top Level](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).

another tried will be pass i18n object to `getArticle` which is what I want to avoid when working on Footer section

Opinion?

src/hooks/useArticle.js
hueitan: same purpose and this addition more-information is used in article page.

src/hooks/useArticle.js
hueitan: done in 4d9394b1c62d344592347a068b2e0ee8edc62163

_not perfect ☹️ we should think more how we parse the html content in general, including translation and markup, when having more idea_

src/api/getArticle.js
```diff
@@ -1,7 +1,14 @@
 import { cachedFetch, buildPcsUrl, canonicalizeTitle } from 'utils'
 
-export const getArticle = (lang, title) => {
+export const getArticle = (lang, title, i18n) => {
   const url = buildPcsUrl(lang, title, 'mobile-sections')
+
+  // retrieve content language and set it back to app language
```
medied: I wonder how often we are going to have to do this `locale` quick switch to grab a label, perhaps a good candidate for a `utils` helper in the future?

src/api/getArticle.js
```diff
@@ -85,6 +93,31 @@ const fixImageUrl = (htmlString) => {
   return htmlString.replace(/src="\/\//gi, 'src="https://')
 }
 
+const fixTableCaption = (htmlString, moreInformationText) => {
+  const hiddenClassName = 'hidden-in-table'
+  const parser = new DOMParser()
+  const node = parser.parseFromString(htmlString, 'text/html')
+  const tableNodes = node.querySelectorAll('table.wikitable, table.multicol')
+  for (const tableNode of tableNodes) {
+    const thContent = Array.from(tableNode.querySelectorAll('th')).map(th => th.textContent).join(', ')
```
medied: Is this `th` as in table header content?

medied:
The tables I was able to find in some articles ([like this one](http://localhost:8080/#/article/en/Coronavirus_disease_2019/Signs_and_symptoms)) look much better than I was expecting, well done Huei. 

What's an article with a big hairy table I can test? I couldn't find the multilingualism one depicted in the PR description

src/api/getArticle.js
hueitan: yeah, I like the idea, banana doesn't support this, we can do our own version of getting translation once in the other language

src/api/getArticle.js
hueitan: yes

hueitan:
> The tables I was able to find in some articles ([like this one](http://localhost:8080/#/article/en/Coronavirus_disease_2019/Signs_and_symptoms)) look much better than I was expecting, well done Huei.
> 
> What's an article with a big hairy table I can test? I couldn't find the multilingualism one depicted in the PR description

try this nice example of article, with some inline images 😃 
http://localhost:8080/#/article/en/List_of_Indian_state_symbols

src/api/getArticle.js
hueitan: Currently, `useI18n()` can also be used in the component level, we need a way to make `banana` available in utils as well, let's do it in another refactor pr including the change for search language and article footer.

medied: I can create Phabricator ticket

medied:
> http://localhost:8080/#/article/en/List_of_Indian_state_symbols
> http://localhost:8080/#/article/en/Population_of_Canada/17th_century

Yea these look good. Seems some tables are going to look better than others just depending on the content. Indian state symbols table looks better to my eyes than population of Canada one because of the amount of text per cell, for example. 

Mango is a good fruit to have as state fruit, also

src/api/getArticle.js
medied: https://phabricator.wikimedia.org/T248354

stephanebisson: The banana instance is a stateful and expensive object so it is made available through a context. That makes it unsuitable for utils. On the other hand I agree that the way we use it with the content language should be refactored. We could do something like `const i18nInContentLang = useI18n(article.contentLang)`

Building the UI with localized strings should stay in components. In this case, user-created article content needs to be modified and cached in the api's transform function. I would prefer that `useArticle` passes only the needed string(s) to `getArticle` instead of the banana instance.

hueitan: Done passing the translation instead of i18n here bdfe89b4ccc782aebffc7410ec413d4c06342f80

stephanebisson:
This has been approved for a few days. @jpita please consider that a priority.

jpita:
http://localhost:8080/#/article/en/Enteritis

on this article, the table (in wikitext, it's a table, not a wikitable) shows and is selectable but has no text/title so the user will miss it
image

hueitan:
Good catch @jpita ,it's fixed now.

hueitan:
FYI, relevant ticket but not going to fix in this pr, https://phabricator.wikimedia.org/T248716

PR #187, opened 2020-03-20
T236318 Add About Wikipedia Page

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T236318

### Note

Reminder that it copies all the image and text from Onboarding page due to the similar content but different UI

Several things can be refactored but can wait

1. Header for the page can be generalized so we can reuse it in component and css
1. Let About App and About Wikipedia open as an popup, so user have the previous settings item selected when Close the about page
1. Reuse the same component for Onboarding and About Wikipedia?

| 1 | 2 | 3 |
| -- | -- | -- |
|  |  |  |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,3 +16,9 @@ jobs:
             ./scripts/zip.sh
       - store_artifacts:
           path: app.zip
+      - store_artifacts:
+          path: cypress/screenshots
+          destination: screenshots
+      - store_artifacts:
+          path: cypress/videos
+          destination: videos
--- a/cypress.json
+++ b/cypress.json
@@ -6,5 +6,5 @@
     "modifyObstructiveCode": false,
     "baseUrl": "http://127.0.0.1:8080",
     "defaultCommandTimeout": 6000,
-    "video": false
+    "video": true
 }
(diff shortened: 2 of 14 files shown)


jpita:
@hueitan we need a "close" button in the last page otherwise in order to leave the about page the user needs to go all the way back to page 1

medied:
Regarding your refactoring ideas:

> Header for the page can be generalized so we can reuse it in component and css

Didn't quite get this, can you explain? Do you mean like a `Header` component to use across different components (`AboutApp`, `AboutWikipedia`, etc)?


> Let About App and About Wikipedia open as an popup, so user have the previous settings item selected when Close the about page

Agree with Pita it would be nice to let the user exit `AboutWikipedia` from the last screen. What about linking a close button at the end to a `route('/settings')` if we don't have time to refactor to opening as popup?


> Reuse the same component for Onboarding and About Wikipedia?

Seems like a good idea, could be a Phabricator task on its own

hueitan:
> Didn't quite get this, can you explain? Do you mean like a `Header` component to use across different components (`AboutApp`, `AboutWikipedia`, etc)?

It can be a `<Header/>` component or just the header stay in popup component, the main purpose is to reuse the component so we don't need to apply the same to every component when header is needed.

> Agree with Pita it would be nice to let the user exit `AboutWikipedia` from the last screen. What about linking a close button at the end to a `route('/settings')` if we don't have time to refactor to opening as popup?

Let's wait for design feedback.

> > Reuse the same component for Onboarding and About Wikipedia?
> 
> Seems like a good idea, could be a Phabricator task on its own

Seems the softkey is different, UI has a bit different, I don't have any good idea now

hueitan:
Confirm and done adding center exit key @medied @jpita 
https://phabricator.wikimedia.org/T236318#5994143

PR #188, opened 2020-03-21
T248080 search language

medied:
Phabricator Link: https://phabricator.wikimedia.org/T248080

---

### Problem Statement

In order to improve search experience, [T248080](https://phabricator.wikimedia.org/T248080) is a short term solution that arose from [T244401](https://phabricator.wikimedia.org/T244401). Main changes are decoupling search language from app language, and bringing `Language` option to `Search` home screen new design from the `Settings` menu. 

### Solution

Proposing to lean on `localStorage` to decouple search language from app language, following design spec.

---

### TODO

- [x] fix select button feedback render in `Language`
- [x] update text from the new design


src/components/Search.js
```diff
@@ -54,9 +59,10 @@ export const Search = () => {
   useSoftkey('Search', {
     right: i18n.i18n('softkey-settings'),
     onKeyRight: () => { window.location.hash = '/settings' },
-    center: current.type === 'DIV' ? i18n.i18n('centerkey-select') : '',
+    center: current.type === 'DIV' ? i18n.i18n('centerkey-select') : i18n.i18n('softkey-search'),
```
medied: [Waiting for confirmation](https://phabricator.wikimedia.org/T248080#5988593) on whether pressing this softkey should make something happen

src/components/Search.js
```diff
@@ -39,10 +40,14 @@ export const Search = () => {
     }
   }
 
-  const goToRandomArticle = () => {
-    getRandomArticleTitle(lang).then(title => {
-      goto.article(lang, title)
-    })
+  // const goToRandomArticle = () => {
```
medied: We've discussed we don't want to remove the random article feature altogether, there's a ticket already ([T248124](https://phabricator.wikimedia.org/T248124)) to decide where we should move it to for testing purposes. 

On that note, I think once we merge this PR we will have enough to mark ([T245156](https://phabricator.wikimedia.org/T245156)) as resolved. 

To be decided if we leave these respective lines commented out in this PR or deleted

src/components/Language.js
```diff
@@ -19,9 +19,10 @@ export const Language = () => {
       const itemIndex = index - 1
       const item = items[itemIndex]
 
-      i18n.setLocale(item.lang)
+      // i18n.setLocale(item.lang)
       setLang(i18n.locale)
-      setAppLanguage(item.lang)
+      // setAppLanguage(item.lang)
```
medied: [Waiting for confirmation](https://phabricator.wikimedia.org/T248080#5988590) that we'd be removing the ability for users to change app language altogether, that would mean some lines in `Language.js` get deleted

src/utils/languages.js
```diff
@@ -1486,3 +1486,11 @@ export const setAppLanguage = lang => {
 export const getAppLanguage = () => {
   return localStorage.getItem('language-app')
 }
+
+export const setSearchLanguage = lang => {
+  localStorage.setItem('language-search', lang)
+}
+
+export const getSearchLanguage = () => {
+  return localStorage.getItem('language-search')
```
hueitan:
```suggestion
  return localStorage.getItem('language-search') || 'en'
```

default search english when there cannot found from localStorage?

src/components/Search.js
```diff
@@ -69,6 +75,11 @@ export const Search = () => {
   return (
     <div class='search'>
       <img class='double-u' src='images/w.svg' style={{ display: ((searchResults || !isOnline) ? 'none' : 'block') }} />
+      <div class='home-text-box' style={{ display: ((searchResults || !isOnline) ? 'none' : 'block') }}>
+        <div class='welcome-text'>
+          {i18n.i18n('welcome-search-text')}
+        </div>
+      </div>
       <input ref={inputRef} type='text' placeholder={i18n.i18n('search-placeholder')} value={query} onInput={onInput} data-selectable />
```
hueitan: the placeholder should be in search language.

medied: Confirming with design: https://phabricator.wikimedia.org/T248080#5993467

src/components/Search.js
medied: Decided during daily meeting to just remove random article feature separately: https://github.com/wikimedia/wikipedia-kaios/pull/189

src/components/Search.js
medied: Confirming there's no action attached to it

src/utils/languages.js
medied: I like this idea. I had done this in `Search` thinking along the same lines (mimicking `src/index`):

```
  const lang = getSearchLanguage() || navigator.language.substr(0, 2)
```

Do you think it makes sense to have both?

hueitan: Nope, as `getSearchLanguage()` already returns value, there is no way to get the second value.

style/listview.less
```diff
@@ -77,5 +77,28 @@
...
+			font-size: 14px;
+			font-weight: normal;
+			font-stretch: normal;
+			font-style: normal;
+			line-height: normal;
+			letter-spacing: normal;
+			text-align: center;
+			color: #72777d;
+			padding: 7px 35px 0 35px;
+		}
```
hueitan: The styling breaks in landscape mode, and the case when there is no explainer [1]

[1]



---

✨ The initial implementation of the `ListView` and `RadioListView` are designed for the general purpose used, therefore it accepts `{empty}` be text or any component for any customized empty content. That's fine for now as we can work for another refactor pr and also let Radio be within `ListView`

stephanebisson:
The phabricator task is still in "needs design". Let's not get too much ahead of ourselves.

src/utils/languages.js
medied: Will opt for your suggestion then and get rid of `|| navigator.language.substr(0, 2)`

style/listview.less
medied: > the case when there is no explainer [1]

Huei could you please list out steps to get to that state? I'm not able to replicate

src/utils/languages.js
medied: And actually before committing: why wouldn't we be doing the same thing with `getAppLanguage`? Do we need to worry about staying consistent overall with these methods?

hueitan: yup, i think yes, when we remove the changing for app language page, there is no place to set app language in localStorage

Q : index.js `const lang = getAppLanguage() || navigator.language.substr(0, 2)`

what if there isn't `getAppLanguage` and the navigator language is not supported in our language json (assume the edge case)? should we set default to `en`?

style/listview.less
hueitan: Currently there is only search page uses `ListView` with `empty` message key, so when I remove `emptySubText` prop in Search page, then I got the above screenshot.

medied:
Phabricator task has been [deprioritized and moved to backlog](https://phabricator.wikimedia.org/T248080#5998226), does it make sense to close this PR and reopen later?

hueitan:
ya, I think we can close it until we have clear message on the ticket.

medied:
May the third time be the charm then

PR #190, opened 2020-03-24
T248354 i18n get translation

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T248354

### Problem Statement

banana doesn't allow us to grab another locale translation without changing the locale

### Solution

Extend the translation method in i18n

### Note

`i18n.getTranslation()` : can you suggest a better naming of this?

### DIFF
--- a/src/hooks/useI18n.js
+++ b/src/hooks/useI18n.js
@@ -1,4 +1,17 @@
 import { useContext } from 'preact/hooks'
 import { I18nContext } from 'contexts'
 
-export const useI18n = () => useContext(I18nContext)
+export const useI18n = () => {
+  const i18n = useContext(I18nContext)
+
+  i18n.getTranslation = (locale, key, ...parameters) => {
+    const i18nLocale = i18n.locale
+    i18n.setLocale(locale)
+    const translation = i18n.i18n(key, ...parameters)
+    i18n.setLocale(i18nLocale)
+
+    return translation
+  }
+
+  return i18n
+}
--- a/src/hooks/useArticle.js
+++ b/src/hooks/useArticle.js
@@ -7,12 +7,9 @@ export const useArticle = (lang, title) => {
   const [article, setArticle] = useState(null)
   const i18n = useI18n()
 
-  // retrieve content language and set it back to app language
-  const i18nLocale = i18n.locale
-  i18n.setLocale(lang)
-  const moreInformationText = i18n.i18n('more-information')
+  // compile translation in content language
+  const moreInformationText = i18n.getTranslation(lang, 'more-information')
   const translation = { moreInformationText }
-  i18n.setLocale(i18nLocale)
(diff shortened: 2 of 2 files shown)


hueitan:
@stephanebisson @medied what about this approach solution? although i think this should be natively supported by `banana`.

src/hooks/useI18n.js
```diff
@@ -1,4 +1,17 @@
...
-import { I18nContext } from 'contexts'
-
-export const useI18n = () => useContext(I18nContext)
+import { useContext } from 'preact/hooks'
+import { I18nContext } from 'contexts'
+
+export const useI18n = () => {
+  const i18n = useContext(I18nContext)
+
+  i18n.getTranslation = (locale, key, ...parameters) => {
```
stephanebisson: Modifying an external object like that is generally not recommended but I think you're on to something. What about returning a function instead of an object, and depending on whether `useI18n()` is called with a `lang` argument, the function could switch the locale.

A positive side-effect would be that usage everywhere would look like `i18n('softkey-blah')` instead of `i18n.i18n('softkey-blah')`

medied: Hmm I like where this is going

hueitan: `i18n()` is a good idea! The only concern I have is where the lang parameter stay? The arguments are message, arguments and lang.

hueitan: or maybe `useI18n()` returns two method?

one is `i18n( message, ...parameters )`

another one with lang `i18nGetTranslation( lang, message, ...parameters )`?



OR

we have two versions of `useI18n()`

`useI18n()` and `useExtendedI18n()` to provide two kinds of i18n instance

hueitan: after several trying, we can't use `useI18n( lang )` as there is only one instance, so unable to share between `useI18n()` and `useI18n( lang )`

I will continue try and see what else do we have

welcome your idea here 🍺

stephanebisson: I have not tried it but this is roughly what I meant:

```
export const useI18n = lang => {
	const banana = useContext(I18nContext)

	return (key, ...args) => {
		if (lang) {
			const locale = banana.locale
			banana.setLocale(lang)
			const msg = banana.i18n(key, args)
			banana.setLocale(locale)
			return msg
		}
		return banana.i18n(key, args)
	}
}

```

The shared instance is only modified inside the call to `i18n()` and the `lang` argument is kept in the closure so different calls of `useI18n()` have different languages. You should be able to do
```
const i18nAppLang = useI18n()
const i18nContentLang = useLang(article.contentLang)
```

hueitan:
The app language switcher still available in Setting page, `i18n.locale` `i18n.setLocale` still needed, therefore closing this and make a note on ticket.

PR #193, opened 2020-03-26
T248286: Jio2 'About the app' and 'About Wikipedia' layout fix

medied:
Phabricator Link: https://phabricator.wikimedia.org/T248286

### Problem Statement

'About the app' and 'About Wikipedia' pages are not rendering correctly in Jio2 devices.

### Solution

Minor CSS adjustments 

### Note

~To be tested on Jio2 device by someone who has one, I have requested this on the Phabricator thread. Let's not merge until it's confirmed it looks good on device~

Tested by Sudhanshu on Jio2 device now

### DIFF
--- a/style/aboutwikipedia.less
+++ b/style/aboutwikipedia.less
@@ -30,7 +30,7 @@
 				}
 			}
 
-			@media ( min-width: 250px ) {
+			@media @landscape {
 				background-size: 90px;
 
 				img {
@@ -45,7 +45,7 @@
 			font-size: 22px;
 			line-height: 26px;
 
-			@media ( min-width: 250px ) {
+			@media @landscape {
 				font-size: 20px;
 				line-height: 22px;
 			}
--- a/style/aboutwikipedia.less
+++ b/style/aboutwikipedia.less
@@ -58,7 +58,7 @@
 			margin-top: 5px;
 
 			@media @landscape {
-				line-height: 14px;
+				line-height: 16px;
 				margin-top: 2px;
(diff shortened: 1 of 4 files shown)


style/aboutwikipedia.less
```diff
@@ -44,6 +44,11 @@
 			font-weight: bold;
 			font-size: 22px;
 			line-height: 26px;
+
+			@media ( min-width: 250px ) {
```
hueitan: what do you think we use it for

`@media @portrait` and `@media @landscape`

http://lesscss.org/#escaping

stephanebisson: Using a variable does sound like a good idea to try to limit the spread of magic numbers.

medied: Not a bad idea Huei, I introduced a `@landscape` variable now. Hoping that name won't be confusing (as these phones don't technically "support" landscape and adjust screen if a user turns it 90 degrees, right?).

The only other media query is a `min-width: 200px` used in `onboarding` and `aboutwikipedia`, does it make sense to set that as a `@portrait`variable?

hueitan: `@portrait` is useful! 

Would be more clear if we add the comment to `variable.style` to said the dimension for portrait 240x320 and landscape 320x240?

medied:
@jpita this is ready to be tested, note that [Sudhanshu did test](https://phabricator.wikimedia.org/T248286) what it looks like on Jio2 device.

PR #196, opened 2020-03-31
T248354 update useI18n() interface

hueitan:
Following the PR of https://github.com/wikimedia/wikipedia-kaios/pull/190

Phabricator Link: https://phabricator.wikimedia.org/T248354

### Problem Statement

banana doesn't allow us to grab another locale translation without changing the locale

### Solution

Two interface usage of i18n : `i18n()` for UI lang, `i18n( lang )` for given lang translation

simplify `i18n.i18n()` to just `i18n()`

### Note

<s>Test spec completely fail, but in the meanwhile, this is ready to be reviewed</s>

### DIFF
--- a/src/hooks/useI18n.js
+++ b/src/hooks/useI18n.js
@@ -1,4 +1,19 @@
 import { useContext } from 'preact/hooks'
 import { I18nContext } from 'contexts'
+import { getAppLanguage } from 'utils'
 
-export const useI18n = () => useContext(I18nContext)
+export const useI18n = lang => {
+  const banana = useContext(I18nContext)
+  banana.setLocale(getAppLanguage())
+
+  return (key, ...args) => {
+    if (lang) {
+      const locale = banana.locale
+      banana.setLocale(lang)
+      const msg = banana.i18n(key, args)
+      banana.setLocale(locale)
+      return msg
+    }
+    return banana.i18n(key, ...args)
+  }
+}
--- a/src/hooks/useI18n.js
+++ b/src/hooks/useI18n.js
@@ -8,9 +8,14 @@ export const useI18n = lang => {
   return (key, ...args) => {
     if (lang) {
       const locale = banana.locale
-      banana.setLocale(lang)
-      const msg = banana.i18n(key, ...args)
-      banana.setLocale(locale)
+      let msg
+      try {
+        banana.setLocale(lang)
+      } finally {
+        msg = banana.i18n(key, ...args)
+        banana.setLocale(locale)
+      }
+
       return msg
     }
 
--- a/src/hooks/useI18n.js
+++ b/src/hooks/useI18n.js
@@ -4,16 +4,22 @@ import { getAppLanguage } from 'utils'
 
 export const useI18n = lang => {
   const banana = useContext(I18nContext)
-  banana.setLocale(getAppLanguage())
 
   return (key, ...args) => {
     if (lang) {
       const locale = banana.locale
       banana.setLocale(lang)
-      const msg = banana.i18n(key, args)
+      const msg = banana.i18n(key, ...args)
       banana.setLocale(locale)
       return msg
     }
+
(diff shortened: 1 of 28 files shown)


hueitan:
Oppss,, I was supposed to open it as Draft PR 😮

src/hooks/useI18n.js
```diff
@@ -1,4 +1,19 @@
 import { useContext } from 'preact/hooks'
 import { I18nContext } from 'contexts'
+import { getAppLanguage } from 'utils'
 
-export const useI18n = () => useContext(I18nContext)
+export const useI18n = lang => {
+  const banana = useContext(I18nContext)
+  banana.setLocale(getAppLanguage())
```
hueitan: do we foresee the performance issue when reseting the locale everytime it calls the useI18n?

stephanebisson: Yes, let's not do that.

https://github.com/wikimedia/banana-i18n/blob/7259156f439fefe9e1239888585ecec6a02699ff/src/index.js#L31

src/hooks/useI18n.js
```diff
@@ -1,4 +1,19 @@
...
+
+  return (key, ...args) => {
+    if (lang) {
+      const locale = banana.locale
+      banana.setLocale(lang)
+      const msg = banana.i18n(key, args)
```
stephanebisson: Do you need to spread `args` like you do below?

src/hooks/useI18n.js
```diff
@@ -1,4 +1,19 @@
...
+import { getAppLanguage } from 'utils'
 
-export const useI18n = () => useContext(I18nContext)
+export const useI18n = lang => {
+  const banana = useContext(I18nContext)
+  banana.setLocale(getAppLanguage())
+
+  return (key, ...args) => {
+    if (lang) {
+      const locale = banana.locale
```
stephanebisson: This logic should be in a try/finally block to ensure we leave the instance with the right language.

src/hooks/useI18n.js
hueitan: Update in https://github.com/wikimedia/wikipedia-kaios/pull/196/commits/caa953a7f27a135d9b3cbfa777a8f074a0f5b3ee

It's still needed but only execute when changing the app language

src/hooks/useI18n.js
```diff
@@ -1,4 +1,30 @@
...
+  const banana = useContext(I18nContext)
+
+  return (key, ...args) => {
+    if (lang) {
+      const locale = banana.locale
+      let msg
+      try {
+        banana.setLocale(lang)
+      } finally {
+        msg = banana.i18n(key, ...args)
```
stephanebisson: I think this line should be in the `try` block

hueitan: if this line in `try` block, msg will be undefined, we need to get it as least the fallback lang

stephanebisson: This is not related to fallback. The try/finally is to ensure that if `banana.i18n(...)` throws an exception, the locale in the shared instance is reset to what it should be.

However, re-reading the code, I think the logic can be simplified to something like
```
lang = lang || getAppLanguage()
if (banana.locale !== lang) {
  banana.setLocale(lang)
}
return return banana.i18n(key, ...args)
```

The idea is that every time the function is called, it sets the right locale (if it needs to). Do you see any issue with this approach?

hueitan: it totally makes sense and the code much simpler now

what benefit actually do we get in try/catch form when setLocale? identify the language we are not supported yet?

stephanebisson:
LGTM

I'll merge it after you update the PR description.

hueitan:
Test specs are fixed and ready to be reviewed / tested / merged.

medied:
I'm kind of late to the party here but just reviewed this and wow this is great, I really like this refactor 👍

PR #197, opened 2020-04-01
T247318: back key functionality

medied:
Phabricator Link: https://phabricator.wikimedia.org/T247318

### Problem Statement

Backspace hardware key functionality is missing

### Solution

Introducing `onKeyBackspace` as an extra `useSoftkey` listener in various parts of the app. Goal is to achieve the expected behavior described in Phabricator ticket description, repeated here for consistency:

> * On phones with the candy bar keyboard, pressing the "BACK" hardware key should take the user one level up from their current state. When they reach the Search screen/the app homescreen, "BACK" should exit the app;
> 
> * On the Qwerty keyboard, allow the "BACK" hardware to take the user to one level up. When they reach the Search screen, "BACK" quits the app. The END CALL key should always quit the app

### Note

I have tested this on browser and banana phone only, I'm asking Pita or Sudhanshu to test on Jio2 device


medied:
When you open an `ArticleMenu` option, such as `Gallery` or `QuickFacts`, the popups are not stacking up. My understanding is that the `Galley` popup should stack on top of the `ArticleMenu` popup itself, correct? 

@hueitan I seem to remember we created a bug ticket for that, yea? Was it [T248171](https://phabricator.wikimedia.org/T248171)?

Currently hitting the backspace key on any article menu option takes you back to the article, not the article menu. If the above is indeed a bug, this is the reason why backspace key doesn't take you "one level up" on article menu options. Note this is not the case with `Setting` menu options, which is not a popup

hueitan:
> When you open an `ArticleMenu` option, such as `Gallery` or `QuickFacts`, the popups are not stacking up. My understanding is that the `Galley` popup should stack on top of the `ArticleMenu` popup itself, correct?

The popup item doesn't stack in `ArticleMenu`, that's as per pervious design, nothing being stack up in the Menu.

> @hueitan I seem to remember we created a bug ticket for that, yea? Was it [T248171](https://phabricator.wikimedia.org/T248171)?

Not related

> Currently hitting the backspace key on any article menu option takes you back to the article, not the article menu. If the above is indeed a bug, this is the reason why backspace key doesn't take you "one level up" on article menu options. Note this is not the case with `Setting` menu options, which is not a popup

There is actually a ticket opened for this, by changing the new page to popup in the settings menu.

hueitan:
Note that the backspace action always be either `close` or `history.back()`, do you think it's possible to set the default behaviour in `useSoftkey` or Softkey component

then there is a problem how we get the `close` event here

medied:
>Note that the backspace action always be either close or history.back(), do you think it's possible to set the default behaviour in useSoftkey or Softkey component
>
>(hmmm not always, in search screen, the backspace action should stick with the system action)

Yea I was actually thinking of something like that too, but as you point out it's not always the same action, like in `Search` screen and there's also `handleKeyBackspace` in Article.js to handle the case when a user navigates to another article from within an article. 

> The popup item doesn't stack in ArticleMenu, that's as per pervious design, nothing being stack up in the Menu.

Hmmm yea you're right, I see now that if `stack: true` is not added as option then it doesn't stack. To elaborate on this, is stacking popups something we're trying to move away from? 

Brainstorming out loud here about how we could achieve the behavior of "going one level up/back" when hitting backspace key within article menu, I'm seeing: 

* if we actually add the `stack: true` option when creating popups for article menu items, then `onKeyBackspace: () => close()` actually achieves the behavior we want
* another idea would be to somehow pass in `showArticleMenu` or similar function as a handler to every popup, but I'm running into `no-use-before-define` errors that way

hueitan:
> Yea I was actually thinking of something like that too, but as you point out it's not always the same action, like in `Search` screen and there's also `handleKeyBackspace` in Article.js to handle the case when a user navigates to another article from within an article.

Let's ignore this then, and it becomes more complex if we do this, same question can also apply to Left Softkey.

> * if we actually add the `stack: true` option when creating popups for article menu items, then `onKeyBackspace: () => close()` actually achieves the behavior we want

for example in `<TextSize/>` component from menu, which is not stacked, when press left softkey, then it will close and go back to article page.

if we do the stack for TextSize in menu, we need to clarify whether left softkey keep the same behaviour (then it is `closeAll()`), and the backspace key has different behaviour (which is `close()`)

> * another idea would be to somehow pass in `showArticleMenu` or similar function as a handler to every popup, but I'm running into `no-use-before-define` errors that way

i'm not sure what do you mean here, i try to answer that - every popup has their own close function, unless you are passing the close function from menu, but not really useful here as we have closeAll function

medied:
>i'm not sure what do you mean here

Yea let's ignore this too... I meant pass in a function that would show the article menu _again_, instead of removing the sub menu popup, but that would probably introduce unnecessary complexity. 

>whether left softkey keep the same behaviour (then it is closeAll()), and the backspace key has different behaviour (which is close())

Exactly. I think it's clear to me from the Phabritcator thread that the close left softkey should remain with same behavior, which means `closeAll` fits appropriately in this case. 

I'll go ahead and push a commit that represents the solution stacking popups in the article menu (with the exception of `TextSize`, to leave as is). This commit brings the app to the desired behavior as described by ticket, we can continue reviewing from there

medied:
Great findings Huie, thanks

> seems like things don't work expected in Quickfact

What is the behavior you are seeing? I tested by opening `QuickFacts` from the article menu and actions, and I'm seeing normal behavior in both.

medied:
Ah Huei! What a good find, I'm glad you caught that 😁 . Keep them coming!

src/components/Language.js
```diff
@@ -30,7 +30,8 @@ export const Language = () => {
     center: i18n('centerkey-select'),
     onKeyCenter,
     left: i18n('softkey-done'),
-    onKeyLeft: () => history.back()
+    onKeyLeft: () => history.back(),
+    onKeyBackspace: () => !query ? history.back() : setQuery(query.slice(0, -1))
```
hueitan, line 34: not sure what's the expected behaviour when user press down to hover any item, then press "Back key", 

if press back, delete the latest input char, then your code here is valid; 

but if press back, goes to the previous screen, then we need to change the code here

medied: Good call, fined-tuned and revised now

src/components/ArticleLanguage.js
```diff
@@ -33,14 +33,24 @@ export const ArticleLanguage = ({ lang, title, close, closeAll }) => {
     closeAll()
   }
 
+  const onKeyBackspace = () => {
+    const selectedElement = document.querySelector('[nav-selected=true]')
```
hueitan, line 37: extra bonus here, 
you can find the `current` in `useNavigation`, checkout how it is used in Search component.

medied:
Let's please test the `Gallery` functionality to make sure I resolved the conflicts with master correctly

src/components/ArticleLanguage.js
medied: Oh gotta love hooks

stephanebisson:
Works well on Jio1 as far as I can tell.

It's a bit sad to see those `history.back()` calls all over the place. There is probably a centralized solution but we don't have to find it now.

medied:
One more minor commit to stack `TextSize` in `ArticleMenu`

> I'll go ahead and push a commit that represents the solution stacking popups in the article menu (with the exception of TextSize, to leave as is).

Earlier I had decided to leave it as it was from before, but thinking about this twice now I don't see why not stack it in the article menu. It keeps the article menu experience more consistent

stephanebisson:
(testing slack integration)

stephanebisson:
(testing slack integration - 2)

PR #199, opened 2020-04-02
T248287: trim user input when changing language

medied:
Phabricator Link: https://phabricator.wikimedia.org/T248287

### Problem Statement

Jio2 devices let users select a "predictive text" when typing a language, this adds an extra space at the end of input string

### Solution

Trimming input in both `Language` and `ArticleLanguage`

### Note

Extra space is doesn't affect search in `Search` input, so leaving as is

### DIFF
--- a/src/components/ArticleLanguage.js
+++ b/src/components/ArticleLanguage.js
@@ -33,6 +33,16 @@ export const ArticleLanguage = ({ lang, title, close }) => {
     close()
   }
 
+  const handleInput = (e) => {
+    const inputQuery = e.target.value
+
+    if (e.inputType === 'deleteContentBackward') {
+      return setQuery(inputQuery)
+    }
+
+    return setQuery(inputQuery.trim())
+  }
+
   useSoftkey('ArticleLanguage', {
     left: i18n('softkey-done'),
     onKeyLeft,
@@ -47,7 +57,7 @@ export const ArticleLanguage = ({ lang, title, close }) => {
   }, [])
 
   return <div class='articlelanguage'>
-    <input type='text' placeholder={i18n('search-language-placeholder')} value={query} onInput={(e) => setQuery(e.target.value.trim())} data-selectable />
+    <input type='text' placeholder={i18n('search-language-placeholder')} value={query} onInput={e => handleInput(e)} data-selectable />
     <RadioListView header={i18n('article-language-available', numOfLanglink)} items={items} containerRef={containerRef} empty={i18n('no-result-found')} />
   </div>
 }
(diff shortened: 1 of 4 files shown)


src/components/ArticleLanguage.js
```diff
@@ -47,7 +47,7 @@ export const ArticleLanguage = ({ lang, title, close }) => {
   }, [])
 
   return <div class='articlelanguage'>
-    <input type='text' placeholder={i18n('search-language-placeholder')} value={query} onInput={(e) => setQuery(e.target.value)} data-selectable />
+    <input type='text' placeholder={i18n('search-language-placeholder')} value={query} onInput={(e) => setQuery(e.target.value.trim())} data-selectable />
```
stephanebisson, line 50: Not related to your change: could you take this opportunity to remove the parenthesis around the `e` argument of the onInput callback? We don't have a strict standard but we generally omit it. Note that it also applies to the other file in this commit.

hueitan:
I have this experience before, if you want to search for language “Bahasa Indonesia”, you will see the problem there ;-)

medied:
Interesting edge case you found there Huei.

> Update: to precisely, type "Bahasa I", then remove the "I"

What I'm seeing when user removes 'I' in the case above: 

* the cursor gets shifted one space to the left because of the trimming, which indeed could throw  the user off. 
* all three Bahasa results still show up on screen

I'm pushing a commit with an idea to review and explore: using [`inputType` property](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType) to only trim input when a user is *not* deleting characters. A few points regarding this idea though:

* it seems to handle the 'Bahasa I' case correctly, as well as still handling the initial 'Hindi ' case
* the one imprecision I'm still seeing is if user *types more than one space*, for example 'Hindi  ' (with two spaces), then:
  * correct results still show up
  * but then if user presses back key, results go blank (no trimming)
  * however on the next back key press, results show up again
  * we can see variations of this with any input if user types many blank spaces for some reason
* as far as browser compatibility, `inputType` is supported by Firefox. I need to test this on device. 

I suppose the scenario of a user typing 'Bahasa I' and the removing the last character would be more common than a user typing many blank spaces for some reason, the latter seems even more of an edge case. But still, can we figure out a better way?

hueitan:
hmmm i didn't think that so details, my idea is just **trim when doing the filter**, so we don't modify the user input and don't need to worry how many space there.

stephanebisson:
> hmmm i didn't think that so details, my idea is just **trim when doing the filter**, so we don't modify the user input and don't need to worry how many space there.

Yeah that makes sense to me. Leave the input as is but trim the value used to filter the language list in `useSearchLanguage` and `useSearchArticleLanguage`

medied:
Ah yea much better 👍

PR #200, opened 2020-04-06
App manifest: use only supported locales

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T248675

### Problem Statement

KaiOS has a limited set of supported language. For displaying basic app information in the store, it needs the exact language codes it knows about. For instance, it supports both `fr-CA` and `fr-FR` and needs to see those codes exactly in the manifest. It won't pick up translations simply for `fr`.

### Solution

Generate the manifest locales from the lists of supported locales for KaiOS and Jio and map to the i18n files as needed.

### DIFF
--- a/scripts/manifest.js
+++ b/scripts/manifest.js
@@ -3,8 +3,121 @@ const fs = require('fs')
 
 let manifest = JSON.parse(fs.readFileSync('manifest.webapp'))
 
-// Set app info (name, subtitle, desc) in all languages available
-manifest.locales = {}
+const jioLocales = [
+	'en-US',
+	'hi-IN',
+	'as-IN',
+	'bn-IN',
+	'gu-IN',
+	'kn-IN',
+	'ks-IN',
+	'ml-IN',
+	'mr-IN',
+	'ne-IN',
+	'or-IN',
+	'pa-IN',
+	'sa-IN',
+	'sd-IN', // RTL
+	'ta-IN',
+	'te-IN',
+	'und-bod',
+	'mai-IN',
+	'kok-IN',
+	'mni-IN',
+	'doi-IN',
+	'ur-IN', // RTL
+]
+
+const kaiLocales = [
+	"af-ZA",
+	"ar-SA", // RTL
+	"az-Latn-AZ",
+	"be-BY",
+	"bg-BG",
+	"bn-IN",
+	"bn-BD",
+	"bs-BA",
+	"cs-CZ",
+	"da-DK",
+	"de-DE",
+	"el-GR",
+	"en-GB",
+	"en-US",
+	"es-US",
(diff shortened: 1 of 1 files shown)


scripts/manifest.js
```diff
@@ -3,28 +3,161 @@ const fs = require('fs')
...
+	'gu-IN',
+	'kn-IN',
+	'ks-IN',
+	'ml-IN',
+	'mr-IN',
+	'ne-IN',
+	'or-IN',
+	'pa-IN',
+	'sa-IN',
+	'sd-IN', // RTL
```
medied, line 20: RTL stands for right to left, correct?

medied:
This seems to be working well. Here's the test that I did with the banana phone:

* First, I changed the language of the device (under the KaiOS settings) to Brazilian Portuguese
* Then I deployed this branch onto device with usual `npm run deploy` method
* Upon successful deployment, opened up the Wikipedia app and it was set to Portuguese right from the start (onboarding screens where in Portuguese)

The above makes sense to me and my understanding is that's the behavior we are looking for. At this point I was able to switch the app language through the app normally, and set it back to Portuguese at the end.

After that:

- I switched the language of device (under the KaiOS settings again) to Spanish (Latam). 
- Then (without redeploying) I opened the app again, and the app was still in Portuguese.

This also makes sense to me, I'm assuming that's expected behavior as well.

hueitan:
I tried to run it locally (on my Windows environment), it seems like the translation is not correct.

Is that only Windows env.? how it looks like in Mac env.?

image

stephanebisson:
> Is that only Windows env.? how it looks like in Mac env.?

The encoding is messed up for you. What does it look like when you open `i18n/fr.json`?

medied:
This is what I get from Mac env:

image

hueitan:
it looks good to me in fr.json file

```json
"app-title": "Wikipédia",
"app-subtitle": "L’encyclopédie libre",
"app-description": "Avec l’application libre Wikipédia, cherchez et explorez des informations du monde entier en plus de 300 langues.",
```

hueitan:
> This is what I get from Mac env:
> 

Ooo.. then forget about Windows env.

hueitan:
I can't find the locale in the language we added before for `Bodo` `Goan Konkani` https://github.com/wikimedia/wikipedia-kaios/pull/156, are these languages we can skip?

stephanebisson:
> I can't find the locale in the supported locales, we added them to language file before for `Bodo` `Goan Konkani` #156, are these languages we can skip?

We only want the languages that are supported by KaiOS or Jio. 

This is to display the app info in the store in the same language as the phone. If there's no way to set your phone to Bodo, then there is no need for this translation in the manifest.

PR #201, opened 2020-04-08
T248381: about pages

medied:
Phabricator Link: https://phabricator.wikimedia.org/T248381

### Problem statement

When you close the `AboutApp` and `AboutWikipedia` screens from `Settings` menu, the navigation position is not being kept

### Solution

Render both `AboutApp` and `AboutWikipedia` screens as popups (instead of having a specific route for it)

### DIFF
--- a/src/components/AboutApp.js
+++ b/src/components/AboutApp.js
@@ -2,14 +2,14 @@ import { h } from 'preact'
 import { useI18n, useSoftkey } from 'hooks'
 import { appVersion } from 'utils'
 
-export const AboutApp = () => {
+export const AboutApp = ({ close }) => {
   const i18n = useI18n()
 
   useSoftkey('AboutApp', {
     right: i18n('softkey-read-more'),
     onKeyRight: () => window.open('https://wikimediafoundation.org/'),
     left: i18n('softkey-close'),
-    onKeyLeft: () => history.back()
+    onKeyLeft: close
   }, [])
 
   return (
--- a/src/components/AboutWikipedia.js
+++ b/src/components/AboutWikipedia.js
@@ -1,7 +1,7 @@
 import { h } from 'preact'
 import { useSoftkey, useI18n, useRange } from 'hooks'
 
-export const AboutWikipedia = () => {
+export const AboutWikipedia = ({ close }) => {
   const i18n = useI18n()
   const [currentIndex, prevOnboard, nextOnboard] = useRange(0, 3)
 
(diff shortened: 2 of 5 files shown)


medied:
In order to solve this ticket, and have the navigation be on the last menu item after closing, I’m rendereing the menu items as a popup (instead of having a sepefic route for it), as hinted at by Huei in Phabricator thread. I do that for `AboutApp` and works fine, gets us the behavior we want on `close`. 

Issue is with `AboutWikipedia`, if I follow the same pattern to render as popup, the softkeys are getting messed when closing that `AboutWikipedia` popup. More specifically, the `Settings` softkeys are not being re-render, instead the softkeys from whatever the last screen is from second popup remain there. The difference between `AboutApp` and `AboutWikipedia` is that we’re passing `true` for the `replace` argument in the `useSoftkeys` hook in `AboutWikipedia`, which is needed to update softkeys there as user navigates between different screens. 

What would be the correct way to approach this? I’ve tried these different ideas but no luck:

* occurred to me that the `const [showAboutWikipedia] = usePopup(AboutWikipedia, { mode: 'fullscreen' })` hook declaration should be moved to top of file to abide by the [hook rules](https://reactjs.org/docs/hooks-rules.html)
    * I thought this could be related to issue, but I see no changes in behavior
* another idea was adding to the dependency array of `useSoftkey` hook in `Settings`
    * for example, tried adding `containerRef` and `current` to that dependency array but no luck: when you close the second popup, these variables are not changing from last value so it seems it’s not triggering the softkeys hook to re-render
* tried even adding a `handleClose` method in `AboutWikipedia` where I would try resetting softkeys before executing `close`

Is there something I’m missing with how to use softkeys hook? Or is this a matter of reviewing the `SoftkeyReducer` logic, such as the `pop` case. 

The second commit related to `AboutWikipedia` includes the bug, left it for explaining purposes

hueitan:
none of the approach solution work because `AboutWikipedia` `Onboarding` are the only places we use `replace` for the `SoftkeyReducer`, and thanks to this ticket we found a new bug from [the previous pr](https://github.com/wikimedia/wikipedia-kaios/pull/84) and [another pr](https://github.com/wikimedia/wikipedia-kaios/pull/92) :tada:

Changing the following code will fix all the problems, let me know if more information is needed here
https://github.com/wikimedia/wikipedia-kaios/blob/65d1384d87dcafb6a3bb5b2bdbd92bb45f427132/src/reducers/SoftkeyReducer.js#L8
```js
const { name, counter } = state.current
return { ...state, current: { name, counter, ...action.config } }
```

medied:
Beautiful, works like a charm. You and I need to grab a drink one day and talk about reducers.

Realizing now I didn't actually create this as a draft a PR, but anyways with the latest commits I think this is officially ready for code review now. I have updated PR description now.

Thanks Huei for the help!

hueitan:
@medied do remember the backspace key functionality change here, and then this is almost ready to go.

medied:
Perfect, just updated with a merge with master. Will let you @hueitan go ahead and signal to Pita if you think it's ready for testing. Thanks!

medied:
@jpita this is ready for QA

PR #202, opened 2020-04-08
T248285: allow letter 's', 'd' and 'f' to change text size in Jio2 device

medied:
Phabricator Link: https://phabricator.wikimedia.org/T248285

### Problem Statement

Because of the different keyboard in Jio2 devices (QWERTY), changing text size with 4, 5 and 6 is not intuitive.

### Solution

Adding the following handlers for three letters to also have the same respective behavior:

* 's' to decrease size like 4
* 'd' to set default like 5
* 'f' to increase size like 6

### Note

I was able to test on browser and banana phone only. On browser, both s/d/f letter and 4/5/6 numbers are working, on banana phone the numbers work as usual and the letters don't (as expected). 

We need to test this on Jio2 device now.

Note that handlers are case sensitive. At the moment I don't know if pressing said letters on Jio2 keyboard will give you lowercase or uppercase, I'm assuming lowercase and left it as such in the code. If it's not working on Jio2, this could be the reason why and we can try with uppercase handlers. We could also include both lowercase and uppercase, but I don't think that should be necessary.

### DIFF
--- a/cypress/integration/article-view-spec.js
+++ b/cypress/integration/article-view-spec.js
@@ -112,6 +112,28 @@ describe('Article view', () => {
     articlePage.getArticleText().should('have.attr', 'style', 'font-size: 20px;')
   })
 
+  it('check text size change with letters', () => {
+    goToCatArticle()
+    articlePage.selectOptionFromArticleMenu('Text size')
+    popupPage.getHeader().should('have.text', enJson['header-textsize'])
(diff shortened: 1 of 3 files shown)


medied:
@jpita you wanna help me test this on Jio2 device? 🤓

jpita:
> @jpita you wanna help me test this on Jio2 device? 🤓

tested, works fine.
adding test

jpita:
lol, 4 commits trying to do 1 amend  :D

medied:
Awesome, thanks Pita for testing. If there are no objections from the rest of the dev team, I think you can merge when you think it's ready

PR #203, opened 2020-04-13
T248443: Add object-fit to gallery.less

medied:
Phabricator Link: https://phabricator.wikimedia.org/T248443

### Problem Statement

Request was to render images so as to not have portrait images stretched to fill entire screen, which looks particularly bad in Jio2 devices

### Solution

Looks like `object-fit: contain` works here. This has been tested by Sudhashu and Pita, they both [said](https://phabricator.wikimedia.org/T248443#6050239) it looks good. I actually hadn't used `object-fit` before, [according to caniuse](https://caniuse.com/#search=object-fit) it is supported by Firefox version 36 and above.


medied:
@jpita when I run test locally I'm not getting any errors. I suggest we rerun workflow to confirm issue before diving into it, it looks like I don't have permission to do so yet:

image

hueitan:
In this change, I saw some image with background color as white, and it should be black.
index

---

Note : The image itself is set to background as white, this is for the svg when the background is given transparent.

medied:
Yea, the behavior I'm seeing on device it's that portrait images render with the white background, and landscape images with the black background. On browser, both portrait and landscape images render with black background, which just makes it a bit harder to debug. 

I tested removing `background-color: #fff;` in `.gallery img` and I was able to get black background to show with portrait images on device, so it seems the white background is overriding the black (but why only for portrait images and not landscape? 🤔). I also tested sticking the `object-fit: contain;` under `.img` (instead of `img`) but this gets us stretched portrait images again. 

>Note : The image itself is set to background as white, this is for the svg when the background is given transparent.

Can you provide an example? Trying to understand if this is an edge case. 

Also, should we pull back the phab ticket from QA column for this?

jpita:
I think white looks better than black, fits the rest of the app

hueitan:
> > Note : The image itself is set to background as white, this is for the svg when the background is given transparent.
> 
> Can you provide an example? Trying to understand if this is an edge case.
> 
> Also, should we pull back the phab ticket from QA column for this?

The pr is here https://github.com/wikimedia/wikipedia-kaios/pull/122, and the ticket's comment is here https://phabricator.wikimedia.org/T236302#5844719

Example:  lang `af` article `Religie`, and the second image

medied:
Thanks Huei for that useful context. I can see the value of the black background, it gives the gallery a bit of a special cinema-like feel. 

I’ve been scratching my head at this one for a while though. I can’t quite get the behavior on the browser and on the device to reconcile. I’ve tried a few different things: 

- setting a combination of different properties on flex child item `img`
  - using `align-self` to override the default value [1]
  - setting `flex` [2]
  - using flex-basis [3]
- adjusting padding [4]

In all those case I either end up with images still stretched vertically or horizontally, or the white background on landscape or portrait. Could it be some of the flex properties not being supported or recognized on device? The issue is getting the same for both landscape or portrait: what's working one way is at the sacrifice of the other. 

Booting up the WebIDE I can further confirm that behavior of the `img` size overlapping parent `div`s, here with a red `border` to visualize debugging:


image

I also tried something perhaps a bit more out there [5] with an extra parent `div`, and this was looking okay(ish?) on device but broken on browser. This I'm leaving in sample branch `T248443-images-jio2-gallery-img-idea`

It occurred to me if for some reason we had to move away from `flex` in gallery thumbnail, we could simulate what's being done on desktop Wikipedia using `display: table` instead. But I'd still like to believe there's a way to do this with `flex`. Any pointers or ideas? 

---

[1] - https://stackoverflow.com/questions/37609642/why-does-flexbox-stretch-my-image-rather-than-retaining-aspect-ratio
[2] - https://stackoverflow.com/questions/29467660/how-to-stretch-children-to-fill-cross-axis
[3] - https://stackoverflow.com/questions/34352140/what-are-the-differences-between-flex-basis-and-width
[4] - https://stackoverflow.com/questions/44009015/use-flexbox-and-maintain-a-11-aspect-ratio-even-though-content-is-sized-differe
[5] - https://stackoverflow.com/questions/30330845/flexbox-image-scale-to-height-and-keep-aspect-ratio

jpita:
moving this to code-review until the discussion is over

medied:
After reviewing this further with Stephane, we've decided to separate this specific bug regarding the white/black background into a separate ticket and simply have white backgrounds only for now. We can't really hold the release for something like this. 

I've created the new ticket and dumped all of the information we've gather so far into it: https://phabricator.wikimedia.org/T250642

@jpita moving this back to QA now

PR #204, opened 2020-04-14
Blacklist languages missing font on devices

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T249723

### Problem Statement

Some languages don't have font available on the devices so they show as tofu

### Solution

Blacklist those languages for now.


jpita:
should we ditch this PR and keep only #205 or I should merge both?

hueitan:
This first then #205

PR #206, opened 2020-04-14
Set referer

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T249216

### Problem Statement

KaiOS app API calls appear to have no referer and that throws off the analysis.

### Solution

Force a dummy referer so it's clear that the page views are internally referred (from the app, to the app)


neilpquinn:
Looks good to me!

PR #208, opened 2020-04-15
T249710 Allow app language to always match device language

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T249710

### Problem Statement

When user changes the device language, the app language doesn't change

### Solution

Having two language storage for device and app language, always check whether the device language has changed before rendering the app

### Note

_code looks ugly ☹️ , do you have any idea?_

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,18 +9,13 @@ import { App } from 'components'
 
 import '../style/style.less'
 
-let banana
-
 if (checkHasDeviceLanguageChanged()) {
   setDeviceLanguage()
-  const lang = getDeviceLanguage()
-  setAppLanguage(lang)
-  banana = new Banana(lang)
-} else {
-  const lang = getAppLanguage()
-  banana = new Banana(lang)
+  setAppLanguage(getDeviceLanguage())
 }
 
+const lang = getAppLanguage()
+const banana = new Banana(lang)
 banana.load(loadAllLanguagesMessages())
 
 render(<App i18n={banana} />, document.querySelector('.root'))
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -1,14 +1,25 @@
...
-const banana = new Banana(lang)
+let banana
+
+if (checkHasDeviceLanguageChanged()) {
+  setDeviceLanguage()
+  const lang = getDeviceLanguage()
+  setAppLanguage(lang)
+  banana = new Banana(lang)
+} else {
+  const lang = getAppLanguage()
```
stephanebisson, line 20: You can probably reorder things here to have 2 blocks
1) reset the app language if the device language has changed
2) unconditionally get app language and create the banana instance

src/utils/languages.js
```diff
@@ -1486,3 +1486,20 @@ export const setAppLanguage = lang => {
 export const getAppLanguage = () => {
   return localStorage.getItem('language-app')
 }
+
+export const setDeviceLanguage = () => {
+  const navigatorLang = navigator.language
+  const lang = navigatorLang.substr(0, navigatorLang.indexOf('-'))
```
stephanebisson, line 1492: Those 2 lines are repeated, they can be abstracted as `getCurrentDeviceLanguage`

stephanebisson:
This doesn't work for me on the device. I think it's because closing and opening the app doesn't really close it. It just goes in the background. I guess the device may decide to kill it sometimes.

We would need to find the right event to run this code.

hueitan:
> This doesn't work for me on the device. I think it's because closing and opening the app doesn't really close it. It just goes in the background. I guess the device may decide to kill it sometimes.
> 
> We would need to find the right event to run this code.

So we compare the device and app language only when boots the app, that's the current change code. We don't do the language switch when the app is opened from the background.

PR #209, opened 2020-04-17
Capture and send back client-side errors

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T248615

### Problem Statement

The KaiOS app may not work as expected on all devices or it may start breaking for any reason.

### Solution

Implement client-side error logging

### Note

To do 
 - [x] Confirm apply this to ALL components or only interested component
    > `<App/>` captures all child components
 - [x] apply this to fetch event
 - [x] where to send app version?
 - [x] provide some secenarios how to reproduce this event
    > Coding Alert 👇 ! Manually add the `throw new Error()` in the child component
```js
// add this to <Search/> Component
if (query && searchResults) {
    throw new Error('error in search component')
}
```


stephanebisson:
FYI when you start sending errors, you can check them on this dashboard: https://logstash.wikimedia.org/app/kibana#/discover/AXGcre7x3bOlOASGBwXV?_g=(refreshInterval%3A(display%3AOff%2Cpause%3A!f%2Cvalue%3A0)%2Ctime%3A(from%3Anow-24h%2Cmode%3Aquick%2Cto%3Anow))

Connect with your wikitech credentials.

package.json
```diff
@@ -70,7 +70,7 @@
   "dependencies": {
     "banana-i18n": "^1.1.2",
     "history": "^4.10.1",
-    "preact": "^10.1.0",
```
hueitan, line 73: we will need this as `useErrorBoundary` start in version `10.2.0`

hueitan:
the intake URL currently returns 404, in the meanwhile the code is ready to be reviewed.

hueitan:
> FYI when you start sending errors, you can check them on this dashboard: https://logstash.wikimedia.org/app/kibana#/discover/AXGcre7x3bOlOASGBwXV?_g=(refreshInterval%3A(display%3AOff%2Cpause%3A!f%2Cvalue%3A0)%2Ctime%3A(from%3Anow-24h%2Cmode%3Aquick%2Cto%3Anow))
> 
> Connect with your wikitech credentials.

Noted for review, I tried to hit the error, but can't see them on the dashboard.

stephanebisson:
Could you investigate if it's possible to include a sourcemap to get a stack trace that makes more sense than that?
```
za@http://localhost:8080/dist/main.js:1:257447
I@http://localhost:8080/dist/main.js:1:187451
O@http://localhost:8080/dist/main.js:1:185156
x/<@http://localhost:8080/dist/main.js:1:181349
x@http://localhost:8080/dist/main.js:1:181246
```

hueitan:
> Could you investigate if it's possible to include a sourcemap to get a stack trace that makes more sense than that?

After several days of trying, I don't think sourcemap (example: [devtool=source-map in weback config](https://webpack.js.org/configuration/devtool/)) can help in the production stack trace, it will be helpful in the development mode and with the modern browser, it gives the source map link in the **developer tools** only)

Annotation 2020-05-07 120654


> za@http://localhost:8080/dist/main.js:1:257447
> I@http://localhost:8080/dist/main.js:1:187451
> O@http://localhost:8080/dist/main.js:1:185156
> x/<@http://localhost:8080/dist/main.js:1:181349
> x@http://localhost:8080/dist/main.js:1:181246

I think what you are recommend here is, we don't want to minify the component/function (za, l, O, x) name here, so we can at least get them when having the error stack trace.

hueitan:
For the latest commit, here's the event we got : `Search`

image

The main.js file size changes from `367782` -> `390486`, it grows a bit

stephanebisson:
@hueitan I updated some of the fields of the error reports. I'll go ahead and merge now but please take a look and let me know if we should change anything.

hueitan:
Looks good to me

jpita:
how can I force an error on the phone to see if this works from the phone?

jpita:
> how can I force an error on the phone to see if this works from the phone?

From Huei's slack response
image

jpita:
tried that on the browser
image

it's not showing anything on kibana
image

this is the code I put inside `export const Search = () => {` just before the `return` 


``` 
if (query && searchResults) {
    throw new Error('error in search component')
  }```

hueitan:
You either run the code in production (app) or you need to allow it by toggle this line https://github.com/wikimedia/wikipedia-kaios/blob/master/src/hooks/useErrorLogging.js#L5

Note that we don't want to send logging error during the development mode.

jpita:
ah! ok, trying now on the device

jpita:
ok, got it https://logstash.wikimedia.org/app/kibana#/doc/logstash-*/logstash-2020.05.12/clienterror?id=AXIIMRbE3bOlOASGWQGP&_g=h@eca78c7

PR #210, opened 2020-04-17
T250253: update useNavigation to support stacked popups

medied:
Phabricator Link: https://phabricator.wikimedia.org/T250253

### Problem Statement

When we changed the article menu so that popups stack, in order to get the desired back key behavior when working on [backspace key feature](https://github.com/wikimedia/wikipedia-kaios/pull/197/files), a bug was introduced because `useNavigation` hook was not prepared to handle a specific case such as `ArticleLanguage` stacked on top of `ArticleMenu`. What was happening is that on every `getAllElements()` call, it was actually getting all elements from both the first and second popup. This was messing up navigation altogether. 

### Solution

At first I was trying to let it grab all elements from both popups and then slicing the array to handle the corresponding elements, but this approach wasn't really working correctly (branch `T250253-article-menu-language-first-approach-idea` for the curious). I opted instead for taking advantage of the `elementsSelector` argument to have the hook grab all the correct elements when `ArticleLanguage` is the `origin`. This required an update to `getSelectedElement()` so that it selects the correct element from the 'active' (latest) popup.

### Note

Still a mystery to me why the first piece of code below works but the second one doesn't

```
allNavSelectedTrue.forEach(element => {
    console.log(‘element.attributes...', element.attributes)
    if (element.getAttribute(currentSelector)) {
       selectedElement = element
    }
})

const selectedElement = Array.from(allNavSelectedTrue).find(element => {
    return element.getAttribute(currentSelector) === true // returns null
})

```

### DIFF
--- a/src/components/ArticleToc.js
+++ b/src/components/ArticleToc.js
@@ -5,9 +5,10 @@ import { ListView } from 'components'
 
 export const ArticleToc = ({ items, currentAnchor, onSelectItem, close, closeAll }) => {
   const containerRef = useRef()
+  const listRef = useRef()
   const i18n = useI18n()
   const listItems = parseTocItems(items)
-  const [, setNavigation, getCurrent] = useNavigation('ArticleToc', containerRef, 'y')
+  const [, setNavigation, getCurrent] = useNavigation('ArticleToc', containerRef, listRef, 'y')
   const onKeyCenter = () => {
     const { index } = getCurrent()
     const item = listItems[index]
@@ -29,8 +30,8 @@ export const ArticleToc = ({ items, currentAnchor, onSelectItem, close, closeAll
     setNavigation(listItems.findIndex(item => item.anchor === currentAnchor))
   }, [])
 
-  return <div class='toc'>
-    <ListView header={i18n('header-sections')} items={listItems} containerRef={containerRef} />
+  return <div class='toc' ref={containerRef}>
+    <ListView header={i18n('header-sections')} items={listItems} containerRef={listRef} />
   </div>
 }
 
--- a/src/components/ArticleToc.js
+++ b/src/components/ArticleToc.js
@@ -14,7 +14,7 @@ export const ArticleToc = ({ items, currentAnchor, onSelectItem, close, closeAll
 
     if (item && item.title) {
       onSelectItem(item)
-      close()
+      closeAll()
     }
   }
   useSoftkey('ArticleToc', {
--- a/src/hooks/useNavigation.js
+++ b/src/hooks/useNavigation.js
@@ -4,19 +4,18 @@ import { useSoftkey } from 'hooks'
 export const useNavigation = (origin, containerRef, axis, elementsSelector = '[data-selectable]') => {
   const [current, setCurrent] = useState({ type: null, index: 0, key: null })
 
-  const getAllElements = () => document.querySelectorAll(elementsSelector)
+  const getAllElements = () => {
+    if (containerRef.current === undefined) {
+      return document.querySelectorAll(elementsSelector)
+    }
+
+    const grandparentContainer = containerRef.current.parentNode.parentNode
+    return grandparentContainer.querySelectorAll(elementsSelector)
+  }
 
   const getSelectedElement = () => {
-    const allNavSelectedTrue = document.querySelectorAll('[nav-selected=true]')
-    const currentSelector = elementsSelector.slice(1, -1)
-    let selectedElement
-    allNavSelectedTrue.forEach(element => {
-      if (element.getAttribute(currentSelector)) {
-        selectedElement = element
-      }
-    })
-
-    return selectedElement
+    const grandparentContainer = containerRef.current.parentNode.parentNode
+    return grandparentContainer.querySelector('[nav-selected=true]')
   }
 
   const getTheIndexOfTheSelectedElement = () => {
--- a/src/hooks/useNavigation.js
+++ b/src/hooks/useNavigation.js
@@ -7,7 +7,16 @@ export const useNavigation = (origin, containerRef, axis, elementsSelector = '[d
   const getAllElements = () => document.querySelectorAll(elementsSelector)
 
   const getSelectedElement = () => {
-    return document.querySelector('[nav-selected=true]')
+    const allNavSelectedTrue = document.querySelectorAll('[nav-selected=true]')
+    const currentSelector = elementsSelector.slice(1, -1)
+    let selectedElement
+    allNavSelectedTrue.forEach(element => {
+      if (element.getAttribute(currentSelector)) {
+        selectedElement = element
+      }
+    })
+
+    return selectedElement
   }
 
   const getTheIndexOfTheSelectedElement = () => {
@@ -15,7 +24,9 @@ export const useNavigation = (origin, containerRef, axis, elementsSelector = '[d
     return element ? parseInt(element.getAttribute('nav-index')) : 0
   }
 
-  const setNavigation = index => selectElement(getAllElements()[index] || document.body)
+  const setNavigation = index => {
+    selectElement(getAllElements()[index] || document.body)
+  }
 
   const navigatePrevious = () => {
     const allElements = getAllElements()
--- a/src/hooks/useNavigation.js
+++ b/src/hooks/useNavigation.js
@@ -1,22 +1,12 @@
 import { useState, useEffect } from 'preact/hooks'
 import { useSoftkey } from 'hooks'
 
-export const useNavigation = (origin, containerRef, axis, elementsSelector = '[data-selectable]') => {
+export const useNavigation = (origin, containerRef, listRef, axis, elementsSelector = '[data-selectable]') => {
   const [current, setCurrent] = useState({ type: null, index: 0, key: null })
 
-  const getAllElements = () => {
-    if (containerRef.current === undefined) {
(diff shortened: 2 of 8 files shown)


src/hooks/useNavigation.js
```diff
@@ -7,15 +7,26 @@ export const useNavigation = (origin, containerRef, axis, elementsSelector = '[d
   const getAllElements = () => document.querySelectorAll(elementsSelector)
 
   const getSelectedElement = () => {
-    return document.querySelector('[nav-selected=true]')
+    const allNavSelectedTrue = document.querySelectorAll('[nav-selected=true]')
+    const currentSelector = elementsSelector.slice(1, -1)
+    let selectedElement
+    allNavSelectedTrue.forEach(element => {
+      if (element.getAttribute(currentSelector)) {
+        selectedElement = element
```
medied, line 15: I'm assuming there's always a `nav-selected=true` element with the corresponding `currentSelector`, but let's talk if that's not a safe assumption

hueitan:
Haven't looked into the details yet, but when test on search page, navigation breaks.

medied:
What behavior are you seeing? Seems to be working fine on my end

hueitan:
You can follow `useLinksNavigation`, when doing the `querySelectorAll`, add the container instead of document to query elements under the container

https://github.com/wikimedia/wikipedia-kaios/blob/b050fc98560ce0c3a6720e6e0919c1c05132d93d/src/hooks/useArticleLinksNavigation.js#L148

medied:
Good idea. Updated PR to follow suit of this better approach, thanks for the feedback Huei 👍

src/components/ArticleToc.js
```diff
@@ -14,7 +14,7 @@ export const ArticleToc = ({ items, currentAnchor, onSelectItem, close, closeAll
 
     if (item && item.title) {
       onSelectItem(item)
-      close()
+      closeAll()
```
medied, line 17: Little (unrelated) bug I found along the way, a bonus fix

src/hooks/useNavigation.js
```diff
@@ -4,10 +4,18 @@ import { useSoftkey } from 'hooks'
...
+    }
+
+    const grandparentContainer = containerRef.current.parentNode.parentNode
+    return grandparentContainer.querySelectorAll(elementsSelector)
+  }
 
   const getSelectedElement = () => {
-    return document.querySelector('[nav-selected=true]')
+    const grandparentContainer = containerRef.current.parentNode.parentNode
+    return grandparentContainer.querySelector('[nav-selected=true]')
```
hueitan, line 18: great, it works now!

two questions and comments

1. why grandparent, not just parent?
1. This fix. Although we will have the problem again if the grandparent or parent component use `useNavigation`, this is not our case as the container usually set to child components. I think we can ignore this now as this seems not a simple fix.

src/components/ArticleToc.js
stephanebisson: I don't advice doing "drive-by" fixes like that. If you do, make sure to provide steps to reproduce the issue that this is fixing.

The preferred way is to file a task describing the issue and link a one-line PR. Yes it takes longer, but it is so much more safer.

src/hooks/useNavigation.js
```diff
@@ -4,10 +4,18 @@ import { useSoftkey } from 'hooks'
 export const useNavigation = (origin, containerRef, axis, elementsSelector = '[data-selectable]') => {
   const [current, setCurrent] = useState({ type: null, index: 0, key: null })
 
-  const getAllElements = () => document.querySelectorAll(elementsSelector)
+  const getAllElements = () => {
+    if (containerRef.current === undefined) {
```
stephanebisson, line 8: I would like to understand when is `containerRef.current` undefined. I feel like querying `document` is never OK but I might be missing some context here.

src/hooks/useNavigation.js
```diff
@@ -4,10 +4,18 @@ import { useSoftkey } from 'hooks'
...
+      return document.querySelectorAll(elementsSelector)
+    }
+
+    const grandparentContainer = containerRef.current.parentNode.parentNode
```
stephanebisson, line 12: Why do you provide a containerRef that is actually 2 levels down from the node we actually need. Can you explain how this works?

medied: This is because we need the `input` element to also be part of the elements being returned by `getAllElements`. The `input` is also `data-selectable` and there are at least three cases where the input is found two level away from where `ref={containerRef}` is being set: [ArticleLanguage.js#L59](https://github.com/wikimedia/wikipedia-kaios/blob/master/src/components/ArticleLanguage.js#L59), [Language.js#L51](https://github.com/wikimedia/wikipedia-kaios/blob/master/src/components/Language.js#L51), [Search.js#L64](https://github.com/wikimedia/wikipedia-kaios/blob/master/src/components/Search.js#L64)

The idea then is to make the scope of the container big enough to contain all needed elements but small enough to also ignore other elements when popups are stacked. I tried a couple different approaches for this

* For example moving the `ref={containterRef}` placement on level up in RadioListView and ListView, and then querying the `containerRef` parent node (instead of grandparent). This query does include the needed `input` and makes navigation work, but it messes up the screen scrolling code which also uses `containerRef` ([useNavigation.js#L80](https://github.com/wikimedia/wikipedia-kaios/blob/master/src/hooks/useNavigation.js#L80)). Scrolling code would have to be updated too.
* Also tried setting `ref={containterRef}` actually two levels out, which means taking it outside of the RadioViewList/ListView component and into Search/Language/ArticleLanguage. But again this meant even further changes needed to get navigation to work. 

So ultimately I’m proposing to grab the grandparent right from within `getAllElements` because it seems to be the least intrusive change that gets us the working behavior, correctly supporting navigation with stacked popups. Does this make sense? Can we think of a better way?

src/hooks/useNavigation.js
medied: `containerRef.current` is undefined when the app first loads the `Search` home screen. If we comment the if-statement here in line 8, this is stack trace that the console throws:

image

It’s from the `setNavigation` in the component's `useEffect`. This error messes up the loading of `Search` because softkeys don’t get set, the cursor doesn’t get set, and querying doesn’t work. 

I am returning `document.querySelectorAll(elementsSelector)` in this case because querying `document` is what we were doing previously.

src/hooks/useNavigation.js
medied: In the case of `getSelectedElement`, we can actully get a working behavior if we just `return containerRef.current.querySelector('[nav-selected=true]’)` directly instead. The only issue with that is a harmless `TypeError` if you click on (invisible) center softkey when input element is active with a ListView/RadioListView below:

image

src/hooks/useNavigation.js
stephanebisson: Thanks for the explanation. I remember now the issue with the input being conceptually part of the list for navigation but not for "scroll into view".

If there's a simple way to capture and use 2 refs, one for each purpose, I think that would be more clear but if not, a prominent comment explaining what's going on could do.

src/hooks/useNavigation.js
stephanebisson: `useEffect` runs after the component has rendered and the refs are populated. However, in this case the ref is on the results list which doesn't exist until there are search results. What we are trying to do here with `setNavigation(0)` is to focus the input box. 

Have 2 refs would fix that too. Or we could be more explicit with something along the lines of `inputRef.current.focus()`

src/components/ArticleToc.js
medied: Makes sense

> to reproduce the issue that this is fixing.

Issue was that when if you openned `ArticleToc` through the article menu, selecting a section there would route to correct section but the screen you saw was the article menu. This was because not all popups were closing, so this was [introduced recently](https://github.com/wikimedia/wikipedia-kaios/commit/62a2ef6c5dc7ee4ba47d463b6872f7d1d6632e92#diff-4cddd38dee83bd40943da8843dff58db) when we stacked popups from the article menu.

Will make sure to file a task next time.

src/hooks/useNavigation.js
medied: That's a great idea, what about this: https://github.com/wikimedia/wikipedia-kaios/pull/210/commits/95900ec7965529e1e5bc95faea82b7e003ba5ac0

src/hooks/useNavigation.js
medied: > Although we will have the problem again if the grandparent or parent component use useNavigation, this is not our case as the container usually set to child components. I think we can ignore this now as this seems not a simple fix

I think the [latest commit](https://github.com/wikimedia/wikipedia-kaios/pull/210/commits/95900ec7965529e1e5bc95faea82b7e003ba5ac0) addresses this too then, if I'm understanding you correctly

jpita:
found an issue:
1-open "gatwick airport" article in english
2-select the "quick facts" from the quick actions menu
3-select the first image (there's another bug here that I will create a ticket for)
BUG:
If you use the physical back button here, it goes back to the quick facts.
If you use the "close" button from the app, it goes back to the article.

IMO both buttons should have the same behaviour/call the same method.
There's a similar issue [here](https://phabricator.wikimedia.org/T250984)

src/hooks/useNavigation.js
hueitan: yeah, it looks good now 👍

PR #211, opened 2020-04-20
Fix device language getter

hueitan:
A hotfix to fix the getter function in the previous pr https://github.com/wikimedia/wikipedia-kaios/pull/208

Phab Ticket : https://phabricator.wikimedia.org/T249710#6072863


jpita:
can someone put the steps to replicate the issue so I can try it?

jpita:
on this branch, by deleting the property `language-app` and leaving `language-device` as `en` , I still see the issue.
I'm not sure of:

- how the variables are set on the device
- how to replicate the issue on the device
- whenever I change `language-device` to anything else other than `en`, it falls back to `en` (I guess this is because the browser language is `en`

hueitan:
> on this branch, by deleting the property `language-app` and leaving `language-device` as `en` , I still see the issue.

Please ignore this, there isn't a way for a user to do this. The app always has both languages set in local storage.

> * how to replicate the issue on the device

Open the chrome browser, clear all the localStorage, refresh the app and I see error in console.

You need to find a browser that returns the language without `-`, for example `en` instead of `en-US`, then you will see the issue

> * whenever I change `language-device` to anything else other than `en`, it falls back to `en` (I guess this is because the browser language is `en`

This is expected if your device defaults to English language, so the browser returns english.

jpita:
I'm not entirely sure why/how but I still get the issue when I was trying some stuff with the cypress tests
image
the test flow is:
1-go to cat article directly with url (setting the has-onboard-before varible)
2-change the language to portuguese
3-go to sections and select suggested articles

I tried manually both on the browser and device and can't replicate the issue, this just means that somehow the language is still not being set, although the user might not run into the same scenario

hueitan:
did you refresh the browser after the second step?

jpita:
> did you refresh the browser after the second step?

updated the steps, there was information missing

jpita:
@hueitan https://937-212663631-gh.circle-artifacts.com/0/screenshots/experiences.js/Article%20view%20--%20check%20footer.png
if you want to see it, on your machine do
`npx cypress open`
and run only the `experience.js` file.

cypress/integration/experiences.js
```diff
@@ -0,0 +1,32 @@
...
+    articlePage.selectOptionFromActionsMenu('languages')
+    cy.get('input').type('português')
+    cy.get('.description').should('have.text', 'Gato')
+    cy.downArrow().enter()
+    cy.clickDoneButton()
+    articlePage.title().should('have.text', 'Gato')
+  })
+
+  it('check footer', () => {
+    articlePage.selectOptionFromActionsMenu('sections')
```
hueitan, line 22: I just check on cypress, when you have a new `it()` scope, it clears all the local storage item.

Therefore, the following script is needed at the beginning of the scope

```js
cy.setLocalStorage('has-onboard-before', true)
cy.visit('http://127.0.0.1:8080/#/article/en/Cat')
```

hueitan: @jpita does this work for you?

jpita: well, the `has-onboard-before` variable has nothing to do with the language.
shall we ignore this and merge the code?

hueitan: as per the meeting discussion, please ignore this and merge the code without experience.js.

PR #212, opened 2020-04-27
Ecmascript 2020 - Optional Chaining

hueitan:
~_experimental_ ~ _experimental_ ~ _experimental_ ~ _experimental_ ~ _experimental_~ over

Let me know what do you think


hueitan:
do not merge this! receive error on change article language page, actually we got the same error now on master.

`TypeError: Intl.PluralRules is not a constructor`

Thanks @jpita for finding.

medied:
I'll admit I wasn't aware of optional chaining until this PR, and now after reading and learning more about it I can definitely see the appeal to it. Looking at the files change, indeed the new syntax is visually appealing and seems to simplify code. 

Honestly at first a part of me was weary of introducing new features at this stage, but as I started testing this branch and finding no issues I'm now growing to like this. I tested both on browser and banana device and the app was working find, no bugs stood out. 

[Browser compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility) says it's from version 74 for Firefox, shouldn't that be an issue for our app?

hueitan:
> Honestly at first a part of me was weary of introducing new features at this stage, but as I started testing this branch and finding no issues I'm now growing to like this. I tested both on browser and banana device and the app was working find, no bugs stood out.

I agree with that, any dependencies change should be avoided before the launch; Ideally, keeping the dependencies up to date is highly recommend, therefore I'm pushing this branch as the experiment in the beginning then I found nothing breaks. The decision is on us and we will definitely reach this again very soon.
 
The only thing I dislike is this sentence but I don't see a better way right now

```js
const source = item?.srcset && `https:${item.srcset[0].src}`
```

> [Browser compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility) says it's from version 74 for Firefox, shouldn't that be an issue for our app?

Don't worry about that, the plugin `@babel/plugin-proposal-optional-chaining` covers that 🎉

stephanebisson:
Babel is great when it allows us to write modern js for older platforms but I would prefer we stick to stable and accepted features. We're a bit too far ahead here.

https://www.infoq.com/news/2020/04/es2020-features/

hueitan:
Thanks for the great feedback! and sure, we can revisit this again when it is stable

PR #215, opened 2020-05-04
Map kok-IN to gom (Goan Konkani Wikipedia)

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T251791

### Problem Statement

For the Goan Konkani language, Jio uses the kok-IN code but Wikipedia uses the gom code.

### Solution

Map kok-IN to gom internally.

### DIFF
--- a/src/utils/languages.js
+++ b/src/utils/languages.js
@@ -1453,7 +1453,7 @@ const languages = [
   },
   {
     canonical_name: 'Goan Konkani (Devanagari script)',
-    code: 'gom-deva',
+    code: 'gom',
     name: '\u0917\u094B\u0935\u093E \u0915\u094B\u0902\u0915\u0923\u0940'
   }
 ]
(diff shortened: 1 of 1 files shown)


hueitan:
In this case, does it makes sense a user can also search the Goan Konkani by `kok` in language setting? Currently, it can be searched by `gom`.

stephanebisson:
> In this case, does it makes sense a user can also search the Goan Konkani by `kok` in language setting? Currently, it can be searched by `gom`.

I don't know how useful it would be but it wouldn't hurt. However it's not an easy fix so I wouldn't make it a blocker.

medied:
Just curious, how did you find this bug? Just came across it?

stephanebisson:
> Just curious, how did you find this bug? Just came across it?

Runa went through the list of supported languages and identified this issue.

stephanebisson:
@jpita ready for testing

PR #216, opened 2020-05-04
T236317: Allow users to provide feedback from within the app

medied:
Phabricator Links: 
* main one: https://phabricator.wikimedia.org/T236317
* secondary: https://phabricator.wikimedia.org/T253437

### Problem Statement

We would like to have a way for users to provide feedback to the development team via the app. Users can report bugs, issues or general comments.

### Solution

Following the design and acceptance criteria laid out in task description. 

---

TODOs

- [x] backspace key only when `<textarea>` is empty
- [x] input validation
- [x] placeholder text
- [x] add hypelinks following `content-license`
- [x] do i need to update qqq.json as well with `"settings-help-feedback": "Send feeback”`
- [x] add thank you message after successful sending
- [x] fix breaking test case
- [x] link for Terms of survey statement is still missing (from Zeplin) ([update here](https://phabricator.wikimedia.org/T253437))
- [x] polist landscape css
- [x] test on device: how would the scrolling work on device when input is multiple lines long


i18n/en.json
```diff
@@ -69,7 +69,7 @@
     "settings-privacy": "Privacy policy",
     "settings-term": "Terms of use",
     "settings-rate": "Rate the app",
-    "settings-help-feedback": "Send feeback",
+    "settings-help-feedback": "Send feedback",
```
medied, line 72: Thanks

medied:
@stephanebisson @hueitan  I still need to test on device and needs some polishing overall but you can get a head start on taking a look if you want

src/components/Feedback.js
```diff
@@ -0,0 +1,91 @@
...
+  const items = [
+    { text: `<a data-selectable>${i18n('feedback-terms-of-survey')}</a>`, link: 'https://wikimediafoundation.org/' }, // TODO: link not provided yet
+    { text: `<a data-selectable>${i18n('feedback-privacy-policy')}</a>`, link: 'https://foundation.m.wikimedia.org/wiki/Privacy_policy' },
+    { text: `<a data-selectable>${i18n('feedback-terms-of-use')}</a>`, link: 'https://foundation.m.wikimedia.org/wiki/Terms_of_Use/en' }
+  ]
+  const hyperlinks = items.map(i => i.text)
+
+  const onKeyRight = () => {
+    if (message) {
+      sendFeedback(message)
```
hueitan, line 22: multiple event sent after multiple time press of right key, should we lock or delay or throttle the event sent?

src/components/Feedback.js
```diff
@@ -0,0 +1,91 @@
...
+      setValidInput(true)
+    } else {
+      setValidInput(false)
+    }
+  }
+
+  const onKeyBackspace = () => {
+    if (message && getCurrent().type === 'TEXTAREA') {
+      setMessage(message.slice(0, -1))
```
stephanebisson, line 31: This is the default behavior of the backspace key. Do you know why you have to do it manually? I don't think we do in the search input.

src/components/Feedback.js
```diff
@@ -0,0 +1,91 @@
+import { h } from 'preact'
+import { useState, useRef, useEffect } from 'preact/hooks'
+import { useI18n, useSoftkey, useNavigation } from 'hooks'
+import { sendFeedback } from 'utils'
+
+export const Feedback = ({ close }) => {
+  const containerRef = useRef()
+  const i18n = useI18n()
+  const [message, setMessage] = useState()
+  const [validInput, setValidInput] = useState(null)
```
stephanebisson, line 10: Can you explain the meaning of `validInput`? What does it mean when it's `true`, `false`, or `null`.

src/components/Feedback.js
```diff
@@ -0,0 +1,91 @@
...
+      }, 4000)
+      return () => clearTimeout(timer)
+    }
+  }, [validInput])
+
+  return (
+    <div class='feedback' ref={containerRef}>
+      <div class='header'>
+        {i18n('feedback-header')}
+        {validInput === true ? <div class='success'><span>{i18n('feedback-success')}</span></div> : null}
```
hueitan, line 76:
```suggestion
        {validInput && <div class='success'><span>{i18n('feedback-success')}</span></div>}
```

src/components/Feedback.js
```diff
@@ -0,0 +1,91 @@
...
+      </div>
+      <div class='body'>
+        <div class='textarea-box'>
+          <form>
+            <textarea class={validInput === false ? 'border-error' : 'border-regular'} value={message} placeholder={i18n('feedback-placeholder')} onChange={e => handleChange(e.target.value)} data-selectable />
+            {validInput === false ? <div class='error'>{i18n('feedback-error')}</div> : null}
```
hueitan, line 82:
```suggestion
            {!validInput && <div class='error'>{i18n('feedback-error')}</div>}
```

src/components/Feedback.js
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <div class='feedback' ref={containerRef}>
+      <div class='header'>
+        {i18n('feedback-header')}
+        {validInput === true ? <div class='success'><span>{i18n('feedback-success')}</span></div> : null}
+      </div>
+      <div class='body'>
+        <div class='textarea-box'>
+          <form>
+            <textarea class={validInput === false ? 'border-error' : 'border-regular'} value={message} placeholder={i18n('feedback-placeholder')} onChange={e => handleChange(e.target.value)} data-selectable />
```
hueitan, line 81:
```suggestion
            <textarea class={validInput ? 'border-regular' : 'border-error'} value={message} placeholder={i18n('feedback-placeholder')} onChange={e => handleChange(e.target.value)} data-selectable />
```

src/components/Feedback.js
hueitan: perhaps

`true` - (default value) when user type something even delete them.
`false` - only when the message is null after press send

in this case, then I think we don't need null, 
then we can simplify a lot of the Conditional operator (?:) code in the markup.

src/components/Feedback.js
medied: Yea that's not a bad idea. Just curious, I'm not too familiar with `EventLogging` yet, would multiple send events simply mean that we would get repeated events on the kibana dashboard?

i18n/en.json
```diff
@@ -78,6 +78,16 @@
...
     "softkey-read-more": "Read More",
+    "feedback-header": "Send feedback",
+    "feedback-explanation": "We will use the information you give us to help address technical issues and to improve our services, subject to our $1, $2 and $3.",
+    "feedback-placeholder": "Start writing here...",
+    "feedback-success": "Thank you for your feedback",
+    "feedback-error": "Please write feedback before sending",
+    "feedback-terms-of-survey": "Terms of survey statement",
+    "feedback-privacy-policy": "Privacy policy",
+    "feedback-terms-of-use": "Terms of use",
+    "send": "Send",
```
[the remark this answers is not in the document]
medied: Hmmm yeap, and should be `softkey-cancel` right below, good catch

src/components/Feedback.js
stephanebisson: We have to make sure users cannot submit the same feedback more than once by mistake. We could clear the form after the submission.

This feedback is private and won't show on any kibana dashboard. We'll work with Neil to export the feedback to a spreadsheet (or something) regularly for Angie and a few others to consider.

src/components/Feedback.js
medied: The behavior I see without this logic doesn't handle deleting characters, and currently we're doing the same in the `onKeyBackspace` for the input in `Language` and `ArticleLanguage` ([see here)](https://github.com/wikimedia/wikipedia-kaios/blob/master/src/components/Language.js#L29). 

This was introduced when we implemented the backspace key feature to navigate the app. The reason we don't have this in `Search` is because we didn't need to alter the behavior of the backspace key on that page, there we needed the backspace key to exit the app which is the original behavior.

But perhaps there's a simpler way to do this? 🤔

src/components/Feedback.js
medied: Hmm perhaps there's a way to simplify things here, let's hash it out. 

> Can you explain the meaning of validInput? What does it mean when it's true, false, or null.

`validInput` is a simple state variable I'm using to keep track of when to display the `error` and `success` messages when users attempts to send feedback. As it stands right now: 

* `null` means neither `success` nor `error` messages are displayed, this is the original state when first rendered
* `true` means the input was valid (`message` wasn't an empty string) and therefore a `success` message is shown
* `false` means the input was invalid (empty string) and therefore the `error` message is shown

I keep track of any change of `validInput` in the dependency arrays of both `useEffect` hooks, and make use of `setValidInput` to make the messages appear and disappear according to the behavior I understood to aim for: user attempts to send with no text, `error` message appears, on typing something error message disappears, the attempt to send again with text this time, `success` message appears for 4 seconds only, etc. 

So if we try to simplify this to just `true` and `false` (no `null`), the issue I'm seeing is what to check for to know when to display `success` message:

```
{validInput === true ? <div class='success'><span>{i18n('feedback-success')}</span></div> : null}
```

Makes sense? Happy to see a simpler way, I'm realizing now maybe the name `validInput` is confusing

stephanebisson: I think this is confusing because the UX is confusing. 

1. I don't think we should let the users select RSK (send) before there's any text in the textarea. We can easily handle that by looking at `message`. I find the "error" message completely unnecessary.

2. On success, we should direct people away from the form. Filling a form successfully never leads to the same empty form (unless it's a repetitive task by nature). It should lead to a confirmation and then to wherever the user was before (settings menu?). I suggest showing the confirmation message in our regular popup, on top of the feedback form, and going back to /settings when the user OKs the popup.

medied: I gotta agree with you, what you're proposing makes sense

src/components/Feedback.js
hueitan: > clear the form after the submission

This sounds good instead of delay or throttle implementation!

src/components/Feedback.js
hueitan: We can have the useSoftkey like

```js
useSoftkey('Feedback', {
    right: i18n('send'),
    onKeyRight,
    left: i18n('softkey-close'),
    onKeyLeft: close,
    onKeyBackspace: (message && getCurrent().type === 'TEXTAREA') || close,
    onKeyCenter
}, [message, current])
```

not simpler nor pretty, but we don't need to define slice event

src/components/Feedback.js
hueitan: I agree too! That's better in UX! 

one suggestion here if popup shows the confirmation before sending, that would be even better!

soham4abc: > I think this is confusing because the UX is confusing.
> 
> 1. I don't think we should let the users select RSK (send) before there's any text in the textarea. We can easily handle that by looking at `message`. I find the "error" message completely unnecessary.
> 2. On success, we should direct people away from the form. Filling a form successfully never leads to the same empty form (unless it's a repetitive task by nature). It should lead to a confirmation and then to wherever the user was before (settings menu?). I suggest showing the confirmation message in our regular popup, on top of the feedback form, and going back to /settings when the user OKs the popup.

I completely agree that! from a user point of view if the same empty form is displayed again it causes a confusion from the user's point of view!

medied: Updated success flow now according to feedback and design changes

src/components/Feedback.js
```diff
@@ -0,0 +1,88 @@
...
+const SuccessConfirmationPopup = ({ closeAll }) => {
+  const i18n = useI18n()
+
+  useSoftkey('FeedbackSuccessMessage', {
+    center: i18n('softkey-ok'),
+    onKeyCenter: closeAll,
+    onKeyBackspace: closeAll
+  }, [])
+
+  return <div class='success-message'>
```
stephanebisson, line 84: formatting
```
return (
  <Components.../>
)
```

style/feedback.less
```diff
@@ -66,52 +59,32 @@
...
-		background-color: #323232;
-		color: #fff;
+	.header {
+		height: @pageHeaderHeight;
+		max-height: @pageHeaderHeight;
+		background-color: #d8d8d8;
+		box-sizing: border-box;
+		padding: 3px 0;
 		font-size: 17px;
-		font-weight: normal;
```
soham4abc: This looks better now

stephanebisson:
In the browser with Jio2 resolution, it looks a bit tight

Screen Shot 2020-05-19 at 09 53 59

Keep in mind that some translations will end up with longer text so we need some room to spare if we don't want to scroll.

medied:
Alright I reduced some pixels, this is what it looks like on my Firefox

Screen Shot 2020-05-19 at 13 06 05

Maybe you can confirm it looks better on your browser, and either way once this is tested on a Jio2 I can adjust further if needed (we know it has been the case with other previous tasks)

src/components/Feedback.js
```diff
@@ -0,0 +1,90 @@
...
+        </div>
+        <div class='explanation-box'>
+          <p dangerouslySetInnerHTML={{ __html: i18n('feedback-explanation', ...hyperlinks) }}> </p>
+        </div>
+      </div>
+    </div>
+  )
+}
+
+const SuccessConfirmationPopup = ({ closeAll }) => {
```
hueitan, line 75: we have the component `ConfirmDialog` that share the same markup, but require some update to the `ConfirmDialog`

not urgent but we can use this as well in textsize popup and language info popup.

medied:
FYI added a new commit: https://phabricator.wikimedia.org/T236317#6150054

style/feedback.less
```diff
@@ -0,0 +1,103 @@
...
+		height: @pageHeaderHeight;
+		max-height: @pageHeaderHeight;
+		background-color: #d8d8d8;
+		box-sizing: border-box;
+		padding: 3px 0;
+		font-size: 17px;
+		text-align: center;
+	}
+
+	p.preview-text {
```
hueitan, line 89: Annotation 2020-05-20 160931

I saw some weird padding on top of the text in my device for both confirm popup

medied:
I have narrowed down what was causing [the bug](https://phabricator.wikimedia.org/T236317#6151731) that @jpita found: when a user presses either RSK or LSFK with the focus (cursor) still on the `textarea`, what was actually happening on device was that the focus would remain on the `textarea`, the cursor actually stayed blinking in there. 

This is why when either confirmation popup came up, pressing a softkey other than the backspace would work and backspace would break: pressing backspace was deleting characters instead of closing popup (and upon deleting a character, `message` would update and hence the softkeys would update according to the `useSoftkey` hook in `Feedback` component).

A solution I'm proposing is applying a `blur` on the `textarea` right before showing either popup. I have tested on banana device and seems to work fine. Only thing is that with `CancelConfirmationPopup` in specific, if you press backspace key you successfully land on previous page but the `textarea` cursor is still blurred. Note at this point you can refocus the textarea by simply clicking up/down keys. Yet to reset the focus automatically ideally we could make use of the dependency array of `useEffect`, but what variable would we track?

stephanebisson:
> A solution I'm proposing is applying a `blur` on the `textarea` right before showing either popup. I have tested on banana device and seems to work fine. Only thing is that with `CancelConfirmationPopup` in specific, if you press backspace key you successfully land on previous page but the `textarea` cursor is still blurred. Note at this point you can refocus the textarea by simply clicking up/down keys. Yet to reset the focus automatically ideally we could make use of the dependency array of `useEffect`, but what variable would we track?

It would be great to find a generic solution to disable components that are behind a popup and re-enable them (and give them an opportunity to focus an input or something) if needed. Not needed for this task but something to think about.

medied:
Updated with approved language from legal, that checks off the last TODO

jpita:
minor issue: the links under the text box, in order to change the focus, I need to press down instead of right/left like on an article page.
we should maintain the same experience throughout the app.

jpita:
when the phone is offline, when I press send I still get the "Feedback sent" message when in reality no feedback was sent.
shall we show an error instead?

medied:
> when the phone is offline, when I press send I still get the "Feedback sent" message when in reality no feedback was sent.
> shall we show an error instead?

So according to phab update and after talking with Angie regarding feedback component when offline, the last commit presents a user-friendly solution we can use for the time being:

* if user is offline from the time she first lands on feedback screen, app shows offline panel and saves the user the trouble of writing feedback (that she won't be able to send)
* if user becomes offline while writing:
  * the offline panel is shown, as well as the yellow bar offline indicator at the top for a few seconds. This is consistent with the behavior in the search home screen
  * and was able to confirm that whatever had been written until then does not get lost if users stays in the feedback screen. So if the user lost connection only momentarily, she would still be able to continue writing from where she left off and submit afterwards

i18n/qqq.json
```diff
@@ -79,6 +79,16 @@
 	"view-in-browser": "Label for the view in browser link. Shown in the footer",
 	"about-app-message": "Main message in About App page",
 	"softkey-read-more": "Label of the read more softkey",
+	"feedback-header": "Header title shown in Feedback component, selected from Settings menu",
+	"feedback-explanation": "Message displained in Feedback component that explains how feedback information will be used",
```
stephanebisson, line 83: This needs to include explanation of the 2 arguments ($1 and $2)

Check the `confirm-section` msg for an example.

src/components/SearchOfflinePanel.js
```diff
@@ -0,0 +1,14 @@
+import { h } from 'preact'
+import { useI18n } from 'hooks'
+
+export const SearchOfflinePanel = () => {
```
stephanebisson, line 4: Consider renaming things (component, classes, image, message) since this is not specific to search anymore.

PR #217, opened 2020-05-06
Reset event data when session expires

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T251909

### Problem Statement

When a page is open and hidden for a very long time, the session id gets renewed but the basic event data (start time, page view token, etc) are maintained so it is counted as a single event.

### Solution

When the session expires, basic event data is regenerated and it is counted as a new event. 

For instance, you read an article and close your phone, the visibilitychanged event will cause that event to be sent. You open the app the next day. It is still on that same article. You read a bit more of it and navigate away. It will send that new pageview event as part of a new session.


src/hooks/useTracking.js
```diff
@@ -32,103 +28,103 @@ const getUserId = () => {
...
+      page_open_time: Math.round(totalTime),
+      page_visible_time: Math.round(totalTime - msPaused),
+      section_count: sectionCount,
+      opened_section_count: trackingRef.openedSectionCount,
+      page_namespace: namespace,
+      is_main_page: isSearch,
+      is_search_page: isSearch
+      /* eslint-enable camelcase */
     }
-    window.addEventListener('pagehide', sendBeacon)
```
hueitan, line 130: pagehide event not useful here?

hueitan: when detect complete exit the app.

stephanebisson: How do you test it on the device?

hueitan: Previously, I was testing it on WebIDE

and add a special local storage name in the pagehide event, after the app exit, launch it again and see whether this special name exists.

stephanebisson: I couldn't find a case where visibilitychanged was not enough. Please let me know if you find one.

hueitan: Ok, I saw the comment 

```
// Log the event now since we don't know if the user is ever coming back.
```

So it makes sense to send the event when visibility is hidden 👍 no need pagehide then

stephanebisson:
I'm going to take a chance and merge it so that Neil can start seeing some data produced by this code.

PR #218, opened 2020-05-06
T250219: Allow users to change their consent of instrumentation

medied:
Phabricator Link: https://phabricator.wikimedia.org/T250219

### Problem Statement

Introduce an option in the settings menu where users can accept or decline the InukaPageView instrumentation.

### Solution

Following design request laid out in Phab ticket. This broke down to two new components  `PrivacyTerms` and `UsageReportConsent`, the second inside the first within `Settings`. Also a new utility `usageReportConsent`. The sister task can make use of this utility. 

### Note

Sister task info:

https://phabricator.wikimedia.org/T250218
https://github.com/wikimedia/wikipedia-kaios/pull/219

---

TODOs:

- [x] integrate the actual consent state reading and updating logic
- [x] test on banana phone
- [x] update `useTracking` to return early if consent is negative
- [x] fix bug of `UsageReportConsent` actually rendering the current state set in `localStorage`


cypress/integration/settings-spec.js
```diff
@@ -19,17 +19,15 @@ const aboutWikipediaPage = new AboutWikipediaPage()
...
 const languageSettingsPopupEnglishText = enJson['language-setting-message']
 const languageChangeDutchText = nlJson['language-change']
 const settingsMenuListDutchText = [nlJson['settings-language'],
   nlJson['settings-textsize'],
   nlJson['settings-about-wikipedia'],
-  nlJson['settings-privacy'],
-  nlJson['settings-term'],
-  nlJson['settings-about-app']]
+  nlJson['settings-about-app'],
+  enJson['settings-privacy-terms']] // TODO: update to nlJson when translations is available
```
medied, line 30: @jpita the translation for `settings-privacy-terms` is not yet available, I left it here to expect in english because that's what i18n does. Is this fine or is there an actual process to handle something like this?

src/components/UsageReportConsent.js
```diff
@@ -0,0 +1,47 @@
...
+  }, [])
+
+  return (
+    <div class='usageconsent' ref={containerRef}>
+      <div class='header'>{i18n('usage-consent-header')}</div>
+      <div class='body'>
+        <div class='message'>
+          <p> {i18n('usage-consent-explanation')} </p>
+        </div>
+        <RadioListView items={items} containerRef={listRef} darkerSelectedBackgroundColor />
```
stephanebisson, line 43: Why do we have the `darkerSelectedBackgroundColor` attribute here with no value?

style/radiolistview.less
```diff
@@ -49,6 +49,10 @@
 
 			&[ nav-selected=true ] {
 				background-color: #eaf3ff;
+
+				&[ darker-selected-background=true ] {
```
stephanebisson, line 53: Custom attributes for styling are highly unusual. I know `nav-selected` is an exception, maybe not even a valid one but I suggest we don't continue in this direction. The _regular_ way of doing this is with a css class.

src/components/UsageReportConsent.js
```diff
@@ -0,0 +1,47 @@
...
+  const containerRef = useRef()
+  const listRef = useRef()
+  const hasConsented = getConsentStatus()
+  const [items, setItems] = useState([{ title: i18n('usage-consent-yes'), isSelected: hasConsented }, { title: i18n('usage-consent-no'), isSelected: !hasConsented }])
+  const [, setNavigation, getCurrent] = useNavigation('UsageReportConsent', containerRef, listRef, 'y')
+
+  const onKeyCenter = () => {
+    const { index } = getCurrent()
+    setItems(items.map((item, i) => {
+      i === index ? item.isSelected = true : item.isSelected = false
```
stephanebisson, line 18: This can be written as `item.isSelected = (i === index)`. The parens are not even needed but they may add clarity.

src/components/UsageReportConsent.js
```diff
@@ -0,0 +1,47 @@
...
+      return item
+    }))
+    index === 0 ? setConsentStatus(true) : setConsentStatus(false)
```
stephanebisson, line 21: Similarly, this is really `setConsentStatus(index === 0)`

cypress/integration/settings-spec.js
jpita: if u want, create the translation in spanish and change the test to change the language to spanish instead of dutch.
that's one way to not leave behind any todos.
or I can do the same for portuguese if you don't have time.

src/components/PrivacyTerms.js
```diff
@@ -0,0 +1,57 @@
...
+
+  const [, setNavigation, getCurrent] = useNavigation('PrivacyTerms', containerRef, listRef, 'y')
+
+  useEffect(() => {
+    setNavigation(0)
+  }, [])
+
+  const items = [
+    { title: i18n('settings-usage-reports'), action: onUsageReportConsentSelected },
+    { title: i18n('settings-privacy'), link: 'https://foundation.m.wikimedia.org/wiki/Privacy_policy' },
```
stephanebisson, line 46: The privacy policy URL could use the `/${lang}` as well.

cypress/integration/settings-spec.js
medied: Yea that could work, though I thought we are not supposed to change directly any `i18n/*.json` file other than `en.json` and `qqq.json`. But yes will leave this to you once it's in QA if that's cool, that would be helpful and much appreciated

stephanebisson: > ... I thought we are not supposed to change directly any `i18n/*.json` file other than `en.json` and `qqq.json`.

That's right. If you want to contribute translations to have to create an account on translatewiki.net and do it there.

src/components/PrivacyTerms.js
medied: Hmm yea interesting point, what I'm seeing is:

- `https://foundation.m.wikimedia.org/wiki/Privacy_policy` has a map for other privacy policy in other languages
- looks like the privacy policy page in english is `https://foundation.m.wikimedia.org/wiki/Privacy_policy`
  - because `https://foundation.m.wikimedia.org/wiki/Privacy_policy/en` lands in a page with no text
  - same with non-mobile version `https://foundation.wikimedia.org/wiki/Privacy_policy/en`
- seems like other languages are directed to `meta.wikimedia`: `https://meta.wikimedia.org/wiki/Privacy_policy/fr`
- and `https://meta.wikimedia.org/wiki/Privacy_policy/en` actually lands on a page with what seems to be the same content as `https://foundation.m.wikimedia.org/wiki/Privacy_policy`

Should we just simply update the link `https://meta.m.wikimedia.org/wiki/Privacy_policy/${lang}`. This would deviate from the link specified in the Phabricator ticket description, but we would be saving the user the step of choosing their language

style/radiolistview.less
medied: Yea actually I'm curious to understand what's a better way to do this, I struggled a bit with this approach and wasn't completely convinced at the end. 

This is of course to still be able to reuse `RadioListView` in `UsageReportConsent`. Another alternative I didn't explore much is replicating a radio list directly in `UsageReportConsent` because the options there happen to be binary and simple. 

> The regular way of doing this is with a css class.

Could you unpack this?

src/components/PrivacyTerms.js
stephanebisson: Nevermind, let's keep linking to foundation wiki. This is a legal document so we need to be sure where we link to. Foundation wiki is controlled by the foundation but meta wiki is publicly editable.

src/components/UsageReportConsent.js
medied: Followed up in your comment in radiolistview.less, but yes if we do go this way I can include a true value here

style/radiolistview.less
stephanebisson: Before we talk about ways of doing this... why would this listview have a different selected row color than the other listviews in the app? I recommend not doing that and checking with Sudhanshu now or letting it go through design review.

medied: Yea that too! Will mention it to Sudhanshu and update accordingly

src/hooks/useTracking.js
```diff
@@ -74,6 +74,7 @@ export const useTracking = (
   }, [openedSections])
 
   useEffect(() => {
+    if (!getConsentStatus()) return
```
medied, line 77: Note to self: if [PR#217 is merged to master first](https://github.com/wikimedia/wikipedia-kaios/pull/217/commits/f0170f62c9f61121f989aa768a02bc851f0ccbab#diff-d7429549f12cc3c9aa2af3d9d2d16243), revisit where to return early

src/components/PrivacyTerms.js
```diff
@@ -0,0 +1,57 @@
...
+    { title: i18n('settings-term'), link: `https://foundation.m.wikimedia.org/wiki/Terms_of_Use/${lang}` }
+  ]
+
+  return <div class='privacyterms' ref={containerRef}>
```
stephanebisson, line 50: Code style: we're trying to standardize to the following formatting:
```
return (
  <Components here/>
)
```
so that the indentation between the first and second lines makes sense.

src/components/UsageReportConsent.js
```diff
@@ -0,0 +1,47 @@
...
+import { useNavigation, useI18n, useSoftkey } from 'hooks'
+import { RadioListView } from 'components'
+import { getConsentStatus, setConsentStatus } from 'utils'
+
+export const UsageReportConsent = ({ close }) => {
+  const i18n = useI18n()
+  const containerRef = useRef()
+  const listRef = useRef()
+  const hasConsented = getConsentStatus()
+  const [items, setItems] = useState([{ title: i18n('usage-consent-yes'), isSelected: hasConsented }, { title: i18n('usage-consent-no'), isSelected: !hasConsented }])
```
stephanebisson, line 12: Please reformat this object literal on at least 2 lines so it's readable

src/utils/usageReportConsent.js
```diff
@@ -0,0 +1,10 @@
+const KEY = 'usage-data-consent'
+
+export const getConsentStatus = () => {
+  const consent = (localStorage.getItem(KEY) === 'true')
```
stephanebisson, line 4: Consider collapsing those 2 lines.

src/utils/usageReportConsent.js
```diff
@@ -0,0 +1,9 @@
+const KEY = 'usage-data-consent'
+
+export const getConsentStatus = () => {
+  return (localStorage.getItem(KEY) === 'true')
+}
+
+export const setConsentStatus = (consent) => {
```
stephanebisson, line 7: We usually don't put parenthesis around lambda arguments when there is only 1.

src/utils/usageReportConsent.js
```diff
@@ -0,0 +1,9 @@
...
+  localStorage.setItem(KEY, consent)
```
stephanebisson, line 8: Line 4 above expect the stored value to be exactly `true` or `'true'`. However, this function accepts basically anything. In other words, encapsulation is not air tight. One way to enforce a boolean value here is to use `!!consent`. It will always resolve to `true` or `false`.

medied: Yea that makes sense, updated `setConsentStatus` accordingly. I'd say we should just be cautious of not introducing the edge case below in the future:

`const myBool = !!"false";  // == true`
https://stackoverflow.com/questions/263965/how-can-i-convert-a-string-to-boolean-in-javascript

src/components/UsageReportConsent.js
```diff
@@ -0,0 +1,50 @@
...
+    { title: i18n('usage-consent-no'), isSelected: !hasConsented }
+  ])
+  const [, setNavigation, getCurrent] = useNavigation('UsageReportConsent', containerRef, listRef, 'y')
+
+  const onKeyCenter = () => {
+    const { index } = getCurrent()
+    setItems(items.map((item, i) => {
+      item.isSelected = (i === index)
+      return item
+    }))
```
hueitan, line 23: Refactor idea :

the `isSelected` attribute needs to reset every time having an action, maybe this can be wrapped in the listview/radiolistview components, so we don't have to define the settings in all the components.

PR #220, opened 2020-05-09
T250218: Prompt users for consent to enable app instrumentation

medied:
Phabricator Link:  https://phabricator.wikimedia.org/T250218

### Problem Statement

Prompt users for usage data consent the first time users land on homepage after onboarding

### Solution

Following design request laid out in Phab ticket. This introduces one new component `ConsentPopup` and makes use of `usageReportConsent` utility created from sister task. 

### Note

Sister task info:

https://phabricator.wikimedia.org/T250219
https://github.com/wikimedia/wikipedia-kaios/pull/218


i18n/en.json
```diff
@@ -84,5 +84,9 @@
     "usage-consent-header": "Send usage reports",
     "usage-consent-explanation": "Help make the app better by letting us know how you use it. Data collected is anonymous. Read privacy policy for more details.",
     "usage-consent-yes": "Yes",
-    "usage-consent-no": "No"
+    "usage-consent-no": "No",
+    "softkey-yes": "Yes",
```
stephanebisson, line 88: The `yes`/`no` here should be the same keys as in the previous PR.

src/utils/usageReportConsent.js
```diff
@@ -7,3 +7,7 @@ export const getConsentStatus = () => {
 export const setConsentStatus = consent => {
   localStorage.setItem(KEY, !!consent)
 }
+
+export const getConsentConfirmationStatus = () => {
```
stephanebisson, line 11: Please rethink the name of this function. This is a getter for whether instrumentation consent has been answered.

i18n/qqq.json
```diff
@@ -85,5 +85,9 @@
 	"usage-consent-header": "The header title shown in the Send usage reports page",
 	"usage-consent-explanation": "Message displayed in Send usage reports page that explains how the info collected is used",
 	"usage-consent-yes": "Label used to grant usage data consent in Send usage reports page",
-  "usage-consent-no": "Label used to deny usage data consent in Send usage reports page"
+	"usage-consent-no": "Label used to deny usage data consent in Send usage reports page",
+	"softkey-yes": "Label of the Yes softkey",
+	"softkey-no": "Label of the No softkey",
+	"usage-consent-prompt": "Message displayed in usage data prompt popup in Search home screen",
+  "usage-consent-prompt-header": "Usage data prompt popup header"
```
hueitan, line 92: indent here?

src/utils/usageReportConsent.js
```diff
@@ -7,3 +7,7 @@ export const getConsentStatus = () => {
...
+  return (localStorage.getItem(KEY) === null || localStorage.getItem(KEY) === undefined)
```
hueitan, line 12: when there is no setting, `null` is given, do you think that's enough?

after the setting the key will be boolean, so no way to get it undefined.

src/components/ConsentPopup.js
```diff
@@ -0,0 +1,28 @@
...
+      setConsentStatus(response)
+      close()
+    }
+  }
+
+  useSoftkey('ConsentMessage', {
+    center: i18n('softkey-yes'),
+    onKeyCenter: handleConsent(true),
+    left: i18n('softkey-no'),
+    onKeyLeft: handleConsent(false)
```
hueitan, line 19: do we have the conclusion what should the backspace key action here?

src/utils/usageReportConsent.js
medied: Right exactly, that's the idea. When `null` is returned because there is no `usage-data-consent` setting, then the consent popup is shown. I tested this behavior on my banana phone and it worked as expected. The undefined part is there just in case honestly, we could probably do away with it

src/components/ConsentPopup.js
medied: According to the phab descritption: 

> Pressing back on this dialog exists the app.

I tested this behavior on banana phone and app exists as expected.

src/utils/usageReportConsent.js
hueitan: then what about this instead of checking null and undefined.

```
return !localStorage.getItem(KEY)
```

i18n/en.json
medied: Yea we could reuse the same labels here from the previous PR, which are `usage-consent-no` and `usage-consent-yes`. I intentionally created `softkey-yes` and `softkey-no` becuase these new labels are actually used with softkeys only, whereas `usage-consent-no` and `usage-consent-yes` are used in the radio list view. It adds a bit of redundancy and yet brings in some separation of concerns. But happy to update this if that doesn't really matter, please confirm

src/utils/usageReportConsent.js
medied: How about one of these: `hasUserConsented` or `isConsentSet` or `getConsentAnswer`?

There are only [two hard things](https://martinfowler.com/bliki/TwoHardThings.html)...

src/utils/usageReportConsent.js
stephanebisson: I think it's best to be precise here. `getItem` on a key that was never set returns `null`. That's what we should check.

https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem

medied: Anyone in favor of removing `|| localStorage.getItem(KEY) === undefined` from the if-statement? That is, leaving it precisely as:

```
  return (localStorage.getItem(KEY) === null)
```

i18n/en.json
```diff
@@ -84,5 +84,7 @@
     "usage-consent-header": "Send usage reports",
     "usage-consent-explanation": "Help make the app better by letting us know how you use it. Data collected is anonymous. Read privacy policy for more details.",
     "usage-consent-yes": "Yes",
-    "usage-consent-no": "No"
+    "usage-consent-no": "No",
+    "usage-consent-prompt": "Help make the app better by letting us know how you use it. You can change your preference at anytime from the settings menu. Data collected is anonymous.",
```
stephanebisson, line 88: Not sure why this message exists and is different than `usage-consent-explanation`. It looks like it's the same question being asked in 2 different places.

medied:
Good point, testing out on device something like this in Search.js:

```
const onInput = ({ target }) => {
    if (isOnline && !hasConsentBeenAnswered()) {
      setQuery(target.value)
    }
  }
```

i18n/en.json
medied: Will check with design

cypress/integration/article-view-spec.js
```diff
@@ -16,6 +16,7 @@ const searchPage = new SearchPage()
 
 describe('Article view', () => {
   beforeEach(() => {
+    localStorage.setItem('usage-data-consent', false) // Don't display usage data consent prompt
```
hueitan, line 19: let's set this to the command `navigateToHomePage`, similar to `has-onboard-before`, only set it manually when needed

source : https://github.com/wikimedia/wikipedia-kaios/blob/master/cypress/support/commands.js#L53

hueitan:
After I select Yes in the consent popup, then search for something and enter the article, I don't see the tracking code being sent, can you confirm this?

stephanebisson:
> After I select Yes in the consent popup, then search for something and enter the article, I don't see the tracking code being sent, can you confirm this?

Yes, I see this too. We have to update `useTracking` to start collecting but only send if the user accepts.

medied:
Great, just addressed latest feedback and resolved small conflict with master

src/utils/usageReportConsent.js
```diff
@@ -7,3 +7,7 @@ export const getConsentStatus = () => {
 export const setConsentStatus = consent => {
   localStorage.setItem(KEY, !!consent)
 }
+
+export const hasConsentBeenAnswered = () => {
```
stephanebisson, line 11: The logic is reversed between the name and implementation. If the consent has been answered, then it is not null.

src/hooks/useTracking.js
```diff
@@ -114,11 +114,12 @@ export const useTracking = (
       /* eslint-enable camelcase */
     }
 
-    sendEvent(SCHEMA_NAME, SCHEMA_REV, language, event)
+    if (getConsentStatus()) {
+      sendEvent(SCHEMA_NAME, SCHEMA_REV, language, event)
+    }
   }
 
   useEffect(() => {
```
stephanebisson, line 122: We should return early (at the very beginning os `useTracking` if the consent is `no`. There is no point running all the code then.

src/utils/usageReportConsent.js
medied: You know... I actually thought of this too, gotta commit to it what I pick up stuff like that

PR #221, opened 2020-05-11
Make instrumentation configurable at build time and disable in CI

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T248667

### Problem Statement

CI generates a lot of garbage instrumentation data

### Solution

Introduce an env variable to control whether instrumentation is enabled. Make it explicitly enabled when building the official app and disabled when building for CI.

I left it enabled in dev because this data is currently valuable to troubleshoot data quality issues.


hueitan:
> I left it enabled in dev because this data is currently valuable to troubleshoot data quality issues.

So ideally when the time is ready, we should change the variable `INSTRUMENTATION=1` existed only in the `zip.sh` file? as we are not interested in either dev / build mode which for developing purpose.

stephanebisson:
> > I left it enabled in dev because this data is currently valuable to troubleshoot data quality issues.
> 
> So ideally when the time is ready, we should change the variable `INSTRUMENTATION=1` existed only in the `zip.sh` file? as we are not interested in either dev / build mode which for developing purpose.

Probably. We'll let Neil decide whether he wants that data in the future.

PR #224, opened 2020-05-12
Apply image style to figure-inline

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T249105

### Problem Statement

Some images didn't get the proper image treatment since it was applied only to figure.

### Solution

Apply the image treatment to figure-inline as well.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -3,6 +3,7 @@
   "rules": {
     "selector-attribute-quotes": null,
     "no-descending-specificity": null,
-    "declaration-no-important": null
+    "declaration-no-important": null,
+    "selector-type-no-unknown": null
   }
 }
--- a/style/article.less
+++ b/style/article.less
@@ -149,7 +149,8 @@
 					}
 				}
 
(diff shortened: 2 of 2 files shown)


hueitan:
should the width be 135px? similar to other img
Screen Shot 2020-05-12 at 11 14 56

Screen Shot 2020-05-12 at 11 15 29

stephanebisson:
> should the width be 135px? similar to other img
> Screen Shot 2020-05-12 at 11 14 56

It's the `max-height: 90px` that's making it small. If I make it larger it doesn't look like a thumbnail anymore and we run the risk that it doesn't fit into 1 page.

hueitan:
> > should the width be 135px? similar to other img
> > Screen Shot 2020-05-12 at 11 14 56
> 
> It's the `max-height: 90px` that's making it small. If I make it larger it doesn't look like a thumbnail anymore and we run the risk that it doesn't fit into 1 page.

right, it's max-height, looks good, that's expected then.

PR #225, opened 2020-05-12
Style all tables

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T248716

### Problem Statement

Some tables are completely hidden.

### Solution

Style all tables but hide the infobox explicitly.


stephanebisson:
Need to fix the tests

hueitan:
looks good to me! Thanks for making it general!

stephanebisson:
@jpita can you help with troubleshooting the tests? I don't understand what's going on.

jpita:
@stephanebisson in the cat article, when you open the quick facts, nothing is selected/with focus.
I believe that on master the first image is on focus and then we go to the first link navigating with the arrows, which is different on this branch.

stephanebisson:
> @stephanebisson in the cat article, when you open the quick facts, nothing is selected/with focus.
> I believe that on master the first image is on focus and then we go to the first link navigating with the arrows, which is different on this branch.
> 

Good catch! That's exactly it.

style/variables.less
```diff
@@ -13,7 +13,9 @@
 @smallFont: 12px;
 @bigFont: 17px;
 
-@data-selected-selector: &[ data-selected='true'];
+@data-selected-selector: &[data-selected='true'];
```
medied, line 16: A step towards towards [T242364](https://phabricator.wikimedia.org/T242364) 👍

PR #226, opened 2020-05-14
stylelint-config upgrade

soham4abc:

### Problem Statement
stylelint-config upgrade

### Solution
The stylelint-config version is upgraded and the app is tested to make sure it's working smoothly

Screenshot also attached ensuring the stability of the app.
Screenshot (73)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "npm run cypress:lint && npm run cypress:run",
     "build": "webpack --mode production",
-    "dev": "webpack-dev-server --mode development",
+    "dev": "INSTRUMENTATION=1 webpack-dev-server --mode development",
     "start": "http-server -p 8080",
     "banana": "banana-checker i18n/",
     "adb:connect": "adb forward tcp:6000 localfilesystem:/data/local/debugger-socket",
@@ -67,7 +67,7 @@
     "terser-webpack-plugin": "^3.0.1",
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11",
-    "webpack-dev-server": "^3.11.0"
+    "webpack-dev-server": "^3.10.3"
   },
   "dependencies": {
     "banana-i18n": "^1.2.1",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "npm run cypress:lint && npm run cypress:run",
     "build": "webpack --mode production",
-    "dev": "INSTRUMENTATION=1 webpack-dev-server --mode development",
+    "dev": "webpack-dev-server --mode development",
     "start": "http-server -p 8080",
     "banana": "banana-checker i18n/",
     "adb:connect": "adb forward tcp:6000 localfilesystem:/data/local/debugger-socket",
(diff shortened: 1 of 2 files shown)


hueitan:
this script targets the Mac/Linux, if you are running the app on Windows, try remove the `INSTRUMENTATION` and then you don't get the error.

soham4abc:
> this script targets the Mac/Linux, if you are running the app on Windows, try remove the `INSTRUMENTATION` and then you don't get the error.

I did that and now it is running.

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "npm run cypress:lint && npm run cypress:run",
     "build": "webpack --mode production",
-    "dev": "INSTRUMENTATION=1 webpack-dev-server --mode development",
```
stephanebisson, line 9: What is the way to provide environment variables in Windows?

package.json
```diff
@@ -62,12 +62,12 @@
     "less-loader": "^5.0.0",
     "style-loader": "^1.2.0",
     "stylelint": "^12.0.1",
-    "stylelint-config-wikimedia": "^0.8.0",
+    "stylelint-config-wikimedia": "^0.10.1",
```
stephanebisson, line 65: This is a good idea. If you submit a PR just for that I'll accept it.

soham4abc: I would be grateful.I have already submitted the Pull request for this (this string is all about that only) I would be more than happy if you merge it.

package.json
soham4abc: https://docs.oracle.com/en/database/oracle/r-enterprise/1.5.1/oread/creating-and-modifying-environment-variables-on-windows.html#GUID-DD6F9982-60D5-48F6-8270-A27EC53807D0 This link provides all the details you would need to provide environment variables inWindows!

package.json
stephanebisson: This PR has 3 changes. If you create another one with only 1 change: the stylelint config upgrade, I will approve it.

The other points require more discussions so bundling them in the same PR is not ideal.

soham4abc: Okay I am making an PR with only the stylelint-config upgrade other than than that every thing remains same.. right?

package.json
stephanebisson: We want to keep this variable in place for now. You are free to get rid of it in your fork or find a way to make it work on Windows, in which case we could accept your change.

package.json
```diff
@@ -62,12 +62,12 @@
...
     "stylelint-webpack-plugin": "^1.2.3",
     "terser-webpack-plugin": "^3.0.1",
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11",
-    "webpack-dev-server": "^3.10.3"
+    "webpack-dev-server": "^3.11.0"
```
stephanebisson, line 70: What is the reason for upgrading this?

soham4abc: I need to work on this more so that webpack-dev-server can be claimed as upgraded will make a separate PR for that later.

package.json
soham4abc: This variable will remain in place. I figured it out to make it work on windows

package.json
soham4abc: I have made the changes as per your request please see to it!

stephanebisson: Could you update the PR description? It is still about the instrumentation variable.

soham4abc:
Now this PR contains only one change that is the stylelint-config upgrade. Please see to it.

package.json
soham4abc: Yes of course. Doing in a minute!

soham4abc: Done!

stephanebisson:
Thank you for your contribution!

PR #229, opened 2020-05-19
Bring back random in search LSK

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T248124

### Problem Statement

Having the random feature is really useful for testing. It allows going through a lot of articles quickly even on wikis where we can't write the language.

### Solution

Add the random article feature in the search LSK. There is no label but the button is enabled. This is controlled by the "RAND" environment variable. It is enabled in dev and disabled in prod.


src/utils/index.js
```diff
@@ -4,8 +4,8 @@ export * from './articleHistory'
 export * from './cachedFetch'
 export * from './canonicalizeTitle'
 export * from './confirm'
+export * from './featureFlags'
```
soham4abc, line 7: Works well

webpack.config.js
```diff
@@ -37,7 +37,9 @@ module.exports = {
     }),
     new webpack.DefinePlugin({
       APP_VERSION: JSON.stringify(require('./package.json').version),
-      INSTRUMENTATION: JSON.stringify(!!parseInt(process.env.INSTRUMENTATION))
+      INSTRUMENTATION: JSON.stringify(!!parseInt(process.env.INSTRUMENTATION)),
+      // "RANDOM" is already used in the terminal, so using just "RAND"
+      RANDOM: JSON.stringify(!!parseInt(process.env.RAND))
```
jpita, line 42: just nitpicking on naming:
`ARTICLE_RANDOMISER` better/clearer than `RANDOM`?

PR #230, opened 2020-05-20
Update environment variable cross platform settings script

hueitan:
Phabricator Link: -

### Problem Statement

We recently introduce `featureFlags` that allows us to toggle feature between dev and prod app, but the script we use to set the variable is only useful in Mac/Linux platform

### Solution

Ideally we can use `.env` to store the variable, but I think that is overkill right now, can wait until we have more flags or even stable. Therefore use [cross-env](https://www.npmjs.com/package/cross-env) to allow running in cross platform

### Note

I have been tested the script under Windows/Mac platform, please do double check in your platform


hueitan:
@soham4abc if you don't mind, can you also test this in your environment.

soham4abc:
> @soham4abc if you don't mind, can you also test this in your environment.

I will surely test this out on my enviornment . But it will take some time cause we are affected by cyclone and power supply is shut off. I am responding from mobile. Please do wait cause I won't be able to test it as power returns approximately 14 hours from now. Sorry for inconvenience

PR #231, opened 2020-05-20
Update `npm run deploy` script to allow cross platform

hueitan:
Phabricator Link: -

### Problem Statement

In the current statement, `;` is not being recognized by windows ...

### Solution

we can have the seperate `deploy.sh` or `deploy.js` here but again overkill. For now, use the combination of || && to make the statement run in different platform

### Note

tested on Windows/Mac, let me know if it doesn't work.

do merge https://github.com/wikimedia/wikipedia-kaios/pull/230 first before this


package.json
```diff
@@ -6,13 +6,13 @@
...
+    "dev": "cross-env INSTRUMENTATION=1 RAND=1 webpack-dev-server --mode development",
     "start": "http-server -p 8080",
     "banana": "banana-checker i18n/",
     "adb:connect": "adb forward tcp:6000 localfilesystem:/data/local/debugger-socket",
-    "b2g:uninstall": "./node_modules/b2g-cli-tool/b2g.js uninstall Wikipedia",
-    "b2g:install": "./node_modules/b2g-cli-tool/b2g.js install .",
-    "deploy": "INSTRUMENTATION=1 RAND=1 npm run build && npm run adb:connect && ( npm run b2g:uninstall; npm run b2g:install )",
+    "b2g:uninstall": "node ./node_modules/b2g-cli-tool/b2g.js uninstall Wikipedia",
+    "b2g:install": "node ./node_modules/b2g-cli-tool/b2g.js install .",
+    "deploy": "cross-env INSTRUMENTATION=1 RAND=1 npm run build && npm run adb:connect && ( npm run b2g:uninstall || echo ) && npm run b2g:install",
```
hueitan, line 15:
```suggestion
    "deploy": "cross-env INSTRUMENTATION=1 RAND=1 npm run build && npm run adb:connect && npm run b2g:uninstall & npm run b2g:install",
```

we can also do this and let uninstall running in background, but I afraid this will bring some unexpected result.

anyone experience with this?

PR #234, opened 2020-05-25
T253561 consent

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T253561

### Problem Statement

Users need to give consent before they can use the app

### Solution

Always show consent form after onboarding.


src/components/PrivacyTerms.js
```diff
@@ -42,9 +33,8 @@ export const PrivacyTerms = ({ close }) => {
   }, [])
 
   const items = [
-    { title: i18n('settings-usage-reports'), action: onUsageReportConsentSelected },
-    { title: i18n('settings-privacy'), link: 'https://foundation.m.wikimedia.org/wiki/Privacy_policy' },
-    { title: i18n('settings-term'), link: `https://foundation.m.wikimedia.org/wiki/Terms_of_Use/${lang}` }
+    { title: i18n('settings-privacy'), action: goto.privacyPolicy },
+    { title: i18n('settings-term'), action: goto.termsOfUse }
```
hueitan, line 37: Listview read the prop `link` to show the link icon next to each item here, the fastest way to do it is adding `link: true` to each item

src/utils/usageReportConsent.js
```diff
@@ -1,13 +1,14 @@
...
+export const isConsentGranted = () => {
+  return localStorage.getItem(KEY) !== null
 }
 
-export const hasConsentBeenAnswered = () => {
-  return (localStorage.getItem(KEY) !== null)
+export const grantConsent = () => {
+  localStorage.setItem(KEY, JSON.stringify({
+    timestamp: Date.now(),
+    version: appVersion()
```
hueitan, line 12: we set the version but never check it, do we need to show the consent every version update?

stephanebisson: There is a chance we need to send the consent to the server eventually. Setting the timestamp and version makes it very easy for us to deal with consent granted on an older version of the app.

src/components/Routes.js
```diff
@@ -1,19 +1,22 @@
...
-import { Article, Search, Settings, Language, Onboarding } from 'components'
-import { onboarding } from 'utils'
+import { Article, Search, Settings, Language, Onboarding, Consent } from 'components'
+import { onboarding, isConsentGranted, goto } from 'utils'
 
 export const Routes = ({ onRouteChange }) => {
   const onChange = ({ url }) => {
     if (!onboarding.isDone()) {
       route('/onboarding')
+    } else if (!isConsentGranted()) {
```
medied, line 11: Makes sense to abstract this out of `Search.js` 👍

src/utils/goto.js
```diff
@@ -9,4 +9,17 @@ const article = (lang, title, replace = false) => {
   route(`/article/${lang}/${titleStr}`, replace)
 }
 
-export const goto = { article }
+const search = () => route('/')
```
medied, line 12: These seem pretty useful

medied:
Is this PR a good opportunity to drop capital T and update to 'Privacy and terms'?

https://github.com/wikimedia/wikipedia-kaios/blob/master/i18n/en.json#L81

I remember Sudhanshu brought it up during a daily and we agreed to change it, but not sure if there's a ticket filed for that

stephanebisson:
> Is this PR a good opportunity to drop capital T and update to 'Privacy and terms'?
> 
> https://github.com/wikimedia/wikipedia-kaios/blob/master/i18n/en.json#L81
> 
> I remember Sudhanshu brought it up during a daily and we agreed to change it, but not sure if there's a ticket filed for that

Yes, I'm doing it in this PR

i18n/en.json
```diff
@@ -78,13 +78,12 @@
...
-    "settings-privacy-terms": "Privacy and Terms",
-    "privacy-terms-header": "Privacy and Terms",
-    "settings-usage-reports": "Send usage reports",
-    "usage-consent-header": "Send usage reports",
-    "usage-consent-explanation": "Help make the app better by letting us know how you use it. Data collected is anonymous. Read privacy policy for more details.",
-    "usage-consent-yes": "Yes",
-    "usage-consent-no": "No",
-    "usage-consent-prompt": "Help make the app better by letting us know how you use it. You can change your preference at anytime from the settings menu. Data collected is anonymous.",
-    "usage-consent-prompt-header": "Anonymous data"
+    "settings-privacy-terms": "Privacy and terms",
```
medied, line 81: 👍

src/utils/consent.js
```diff
@@ -0,0 +1,25 @@
...
+  const version = appVersion()
+  sendEvent('KaiOSAppConsent', 20122102, getAppLanguage(), { version })
+  localStorage.setItem(KEY, JSON.stringify({
+    timestamp: Date.now(),
+    version: appVersion()
+  }))
+}
+
+// Those messages are not in i18n/*.json because they have a legal
+// value and are translated by legal experts. (more languages to come)
```
medied, line 19: I see, so as of right now: do we know if these message will eventually move to `i18n/*` or are they to stay in this util for the foreseeable future? How would the translation work, we add them directly here once we obtained them from legal experts?

medied:
Latest changes look good, I tested on banana device and all good. With consent messages in English there is still some space in screen, so it seems translations should fit well.

src/utils/consent.js
stephanebisson: They should never be in `i18n/`. We'll receive professional translation for our target languages soon.

I think when it grows I will want to move it somewhere else so it's in its own file and easier to find.

PR #236, opened 2020-05-27
T250642: bring back the gallery black background

medied:
Phabricator Link: https://phabricator.wikimedia.org/T250642

### Problem Statement

We had decided to temporarily remove the black background in the gallery as a consequence of [another gallery-related task](https://phabricator.wikimedia.org/T248443). The goal here is to bring the black background back and have the gallery display it correctly for both landscape and portrait images on device.

### Solution

After many deploys trying out a number of ideas, I was finally able to get the device to render images properly with the use of `onImageLoad` only. The solution consists of looking at the dimensions of each new image in the gallery and deciding based on its width and height whether said image is a portrait or landscape. With this information, we can then set CSS appropriately for each case. Note that:

- You will notice there are three classes in `gallery.less` that pertain to the height/width of the images: `portrait`, `landscape` and `hasHeader`
- In order to obtain an accurate reading of the image's width/height, we have to first remove the classes that are related to that (`portrait`, `landscape` and `hasHeader`)
  - This was probably the hardest to actually realize after all the trial and errors. The behaviors that I see on device if you don't remove the classes first vary and are all ultimately wrong: landscape images off-centered, portrait images with a white background, etc. If we don't clean up the image first (by removing classes), the height/width readings are not accurate and thus the logic falls apart.
- You will see what seems to be a redundant and unnecessary line of code in `gallery.less`: the `height` being set in `.img`. We need it this way for this solution to work because when we remove the classes we must reset height for an accurate reading. 

I encourage you to test this on device and welcome ideas to make it simpler/better.

### DIFF
--- a/src/components/Gallery.js
+++ b/src/components/Gallery.js
@@ -93,8 +93,12 @@ export const Gallery = ({ close, closeAll, items, startFileName, lang }) => {
       galleryNode.classList.remove(galleryClass)
     })
 
-    img.height >= img.width ? galleryNode.classList.add('portrait') : galleryNode.classList.add('landscape')
-    if (items[currentIndex].caption) galleryNode.classList.add('hasHeader')
+    const orientationClass = img.height >= img.width ? 'portrait' : 'landscape'
+    galleryNode.classList.add(orientationClass)
+
+    if (items[currentIndex].caption) {
+      galleryNode.classList.add('hasHeader')
+    }
   }
 
   useSoftkey('Gallery', {
@@ -117,7 +121,7 @@ export const Gallery = ({ close, closeAll, items, startFileName, lang }) => {
         )
       }
       <div class='img'>
-        <img onLoad={e => onImageLoad(e)} src={items[currentIndex].thumbnail} />
+        <img onLoad={onImageLoad} src={items[currentIndex].thumbnail} />
       </div>
     </div>
   )
--- a/src/components/Gallery.js
+++ b/src/components/Gallery.js
@@ -1,5 +1,5 @@
 import { h } from 'preact'
-import { useRef, useLayoutEffect } from 'preact/hooks'
+import { useRef, useLayoutEffect, useState } from 'preact/hooks'
 import { useI18n, useSoftkey, usePopup, useRange, useArticleMediaInfo } from 'hooks'
 
 const MAX_DESCRIPTION_HEIGHT = 45
@@ -83,6 +83,12 @@ export const Gallery = ({ close, closeAll, items, startFileName, lang }) => {
     currentIndex, onPrevImage, onNextImage
   ] = useRange(getInitialIndex(items, startFileName), items.length - 1)
   const [showAboutPopup] = usePopup(AboutContainer, { stack: true })
+  const [isPortrait, setIsPortrait] = useState()
+
+  const onImageLoad = ({ target: img }) => {
+    console.log('onImageLoad - img...', img)
+    img.height >= img.width ? setIsPortrait(true) : setIsPortrait(false) // TODO: if they are equal?
+  }
 
   useSoftkey('Gallery', {
     left: i18n('softkey-close'),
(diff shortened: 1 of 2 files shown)


src/components/Gallery.js
```diff
@@ -79,11 +79,24 @@ const LoadingAbout = () => {
...
 
+  const onImageLoad = ({ target: img }) => {
+    const galleryNode = containerRef.current
+    const galleryClasses = ['hasHeader', 'portrait', 'landscape']
+
+    galleryClasses.forEach(galleryClass => {
+      galleryNode.classList.remove(galleryClass)
+    })
+
+    img.height >= img.width ? galleryNode.classList.add('portrait') : galleryNode.classList.add('landscape')
```
medied, line 96: If height and width are equal I'm setting it as portrait, but I'm actually curious to hear if you think it should be landscape

style/gallery.less
```diff
@@ -18,15 +19,34 @@
 		align-items: center;
 		justify-content: center;
 		flex-direction: column;
-		height: calc( 100vh - @softkeysHeight );
 
 		img {
-			max-width: 100%;
 			background-color: #fff;
```
hueitan, line 24: this should be removed? I can still see some image with #fff background

medied:
>Have you tried setting the state in image load, then update the class list in useLayoutEffect? I'm seeing the blink changes when updating the class list, wondering will useLayoutEffect helps?

Yeap, I have tried a few different combinations (the [second commit](https://github.com/wikimedia/wikipedia-kaios/pull/236/commits/ac48ea2469b37c5219b0bf07a7a649effa995e52) shows the beginning of that approach), some thoughts/notes:

* I tried using `useLayoutEffect` tracking a `isPortrait` in dependency array, so checking for a change in image orientation, and removing/adding classes as needed

* Thing is the timing is tricky. We have to read `img.height` and `img.width` without the effect of class on it, and we can start reading only as soon as image loads. 

* The height being set by `hasHeader` was particularly troublesome because not removing it would get images be 'stuck' with a height ceiling that throws off the calculation.

* I also tried with a second `useLayoutEffect` tracking `currentIndex` to respond earlier
  * this was far from great because you could see the image blow up in size for a fraction of a second before the transition

* I learned an advantage of doing the class setting logic with `onImageLoad` is that we're guaranteed it will run every time. 

  * there are four different states possible in the gallery: portrait without header, portrait with header, landscape without header and landscape with header
  * for any gallery in any article, we can have all four state and we don't know what comes after each transition
  * the issue I was running into with `useLayoutEffect`, then, is that the transition could be from portrait with header to portrait without header, for example, and the class setting logic would fail to update appropriately, or viceversa. 
  * at one point I had two state variables, `isPortrait` (for orientation) and `hasHeader`, and tried using both with `useLayoutEffect` but wasn't able to accommodate for all possible transitions. Maybe there is a way?

style/gallery.less
medied: We need to have the white background underneath images to accommodate for transparent svg images (https://phabricator.wikimedia.org/T236302#5845706)

hueitan:
>   * at one point I had two state variables, `isPortrait` (for orientation) and `hasHeader`, and tried using both with `useLayoutEffect` but wasn't able to accommodate for all possible transitions. Maybe there is a way?

Ya, looks like there isn't a quick fix for this. Once we have this https://phabricator.wikimedia.org/T253466, a loading layer to cover the timing when the layout change.

src/components/Gallery.js
stephanebisson, line 96: Consider switching this around. What varies is the class name, not the call to `classList.add`

```
const orientationClass = img.height >= img.width ? 'portrait' : 'lanscape'
galleryNode.classList.add(orientationClass)
```

src/components/Gallery.js
```diff
@@ -79,11 +79,24 @@ const LoadingAbout = () => {
...
+    if (items[currentIndex].caption) galleryNode.classList.add('hasHeader')
```
stephanebisson, line 97: I know this is not enforced everywhere but please use the form with braces even for one-liner

```
if (...) {
  ...
}
```

src/components/Gallery.js
```diff
@@ -104,7 +117,7 @@ export const Gallery = ({ close, closeAll, items, startFileName, lang }) => {
         )
       }
       <div class='img'>
-        <img src={items[currentIndex].thumbnail} />
+        <img onLoad={e => onImageLoad(e)} src={items[currentIndex].thumbnail} />
```
stephanebisson, line 120: I think this can be simplified to `onLoad={onImageLoad}`

PR #237, opened 2020-05-27
POC - user back to previous screen

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T251517

### Problem Statement

-

### Solution

- [x] Record `Anchor` in the article history and read it when go to the previous article
- [ ] page location
- [ ] is quickfact page

### Note

placeholder, this is POC and ask CircleCI to do the test

### DIFF
--- a/src/components/Article.js
+++ b/src/components/Article.js
@@ -215,13 +215,14 @@ const ArticleInner = ({ lang, articleTitle, initialAnchor }) => {
     if (currentSection !== 0) {
       // lead section doesn't count as opened sections
       setOpenedSections({ ...openedSections, [anchor]: true })
-      goto.article(lang, [articleTitle, anchor], true)
-      articleHistory.update(lang, articleTitle, anchor)
-    } else {
-      goto.article(lang, articleTitle, true)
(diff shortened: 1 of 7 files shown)


hueitan:
Note: I'm still not sure whether we should use `page` state here, it becomes so complex when having more state.

medied:
> Note: I'm still not sure whether we should use `page` state here, it becomes so complex when having more state.

Hmm I think I see what you mean by the complexity, but still trying to follow actually. 

Could you explain what is the difference between `page` and `anchor` in this case? What would be the benefit of tracking `page`?

hueitan:
> Could you explain what is the difference between `page` and `anchor` in this case? What would be the benefit of tracking `page`?

you can notice there are 3 states `title` `anchor` `page`

`anchor` means the section name, example in Cat are `Etymology and naming` `Taxonomy` `Evolution` `Domestication`

`page` means the exact user location, when you press up/down within the article, the page state change.

stephanebisson:
See conclusions in phab: https://phabricator.wikimedia.org/T251517#6247182

PR #238, opened 2020-05-28
T247255: Prioritized Languages

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T247255

### Problem Statement

When the user goes to change the search/UI language, we show a short list of 10 somewhat arbitrary languages.

### Solution

Allow for a special initial list of languages that is more relevant to Jio users in India. 

CI will now generate two application bundles (zip files). One for Jio and one for generic KaiOS.

### DIFF
--- a/scripts/zip.sh
+++ b/scripts/zip.sh
@@ -1,16 +1,19 @@
 #!/bin/bash
 
+TARGET=$1
+FILENAME=WikipediaKaiOS$TARGET
+
 rm -rf Wikipedia
 rm -rf dist
-rm app.zip
+rm $FILENAME.zip
 
 mkdir Wikipedia
 
 node scripts/manifest.js > Wikipedia/manifest.webapp
-cross-env INSTRUMENTATION=1 npm run build
+LANG_LIST=$TARGET INSTRUMENTATION=1 npm run build
 cp -r dist Wikipedia/
 cp -r images Wikipedia/
 cp index.html Wikipedia/
 
 cd Wikipedia
-zip -r ../app.zip *
+zip -r ../$FILENAME.zip *
--- a/src/utils/featureFlags.js
+++ b/src/utils/featureFlags.js
@@ -1,8 +1,11 @@
 // These flags are defined in webpack.config.js
-// and enabled/disabled on the command line
+// and configured on the command line
 
 /* eslint-disable-next-line no-undef */
 export const isInstrumentationEnabled = () => INSTRUMENTATION
 
 /* eslint-disable-next-line no-undef */
 export const isRandomEnabled = () => RANDOM
+
+/* eslint-disable-next-line no-undef */
+export const prioritizedLanguageListName = LANG_LIST
--- a/src/utils/languages.js
+++ b/src/utils/languages.js
@@ -1501,7 +1501,7 @@ const rtl = [
 ]
 
 const prioritizedLists = {
-  default: ['en', 'nl', 'de', 'sv', 'fr', 'it', 'ru', 'es', 'pl', 'pt'],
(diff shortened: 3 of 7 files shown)


src/utils/languages.js
```diff
@@ -1498,11 +1500,33 @@ const rtl = [
...
+  default: ['en', 'nl', 'de', 'sv', 'fr', 'it', 'ru', 'es', 'pl', 'war'],
+  jio: ['en', 'ji', 'as', 'bn', 'gu', 'kn', 'ks', 'ml', 'mr', 'ne', 'or', 'pa', 'sa', 'sd', 'ta', 'te', 'und', 'mai', 'kok', 'mni', 'doi', 'ur']
+}
+
 export const isSupportedForReading = langCode => {
   return missingFont.indexOf(langCode) === -1 &&
     rtl.indexOf(langCode) === -1
 }
 
+export const prioritizedLanguages = languages
```
medied, line 1513: Nice

src/utils/featureFlags.js
```diff
@@ -1,8 +1,11 @@
 // These flags are defined in webpack.config.js
-// and enabled/disabled on the command line
+// and configured on the command line
```
medied, line 2: Could you explain how we can configure on the command line? For example, if I wanted to test 'Jio' language list locally

stephanebisson: You would do `LANG_LIST=jio npm run dev`

soham4abc: works with my environment

scripts/zip.sh
```diff
@@ -1,16 +1,19 @@
...
 rm -rf Wikipedia
 rm -rf dist
-rm app.zip
+rm $FILENAME.zip
 
 mkdir Wikipedia
 
 node scripts/manifest.js > Wikipedia/manifest.webapp
-cross-env INSTRUMENTATION=1 npm run build
+LANG_LIST=$TARGET INSTRUMENTATION=1 npm run build
```
medied, line 13: When `cross-env` was added here it was also added for the `dev` and `deploy` commands in package.json.

https://github.com/wikimedia/wikipedia-kaios/pull/230/files

Should we also update (remove/replace/add) package.json with `LANG_LIST`?

stephanebisson: I've removed `cross-env` from here because it fails for me locally and the primary task of this script is to generate the bundle in CI on linux. No need to be cross env.

I haven't updated the scripts in package.json because I don't this the `jio` lang list should be the default in dev.

hueitan: ya, that's fine. I wasn't able to produce the manifest file correctly even using the cross-env on Windows.

jpita:
@stephanebisson can you send the app to the kaios store to see if everything works well when there?

stephanebisson:
> @stephanebisson can you send the app to the kaios store to see if everything works well when there?

Done

soham4abc:
> One minor comment but overall LGTM, I was able to test on banana phone and desktop successfully

Tested in my environment too! All okay here.

jpita:
> > @stephanebisson can you send the app to the kaios store to see if everything works well when there?
> 
> Done

you deployed the jio one?

stephanebisson:
> > > @stephanebisson can you send the app to the kaios store to see if everything works well when there?
> > 
> > Done
> 
> you deployed the jio one?

We cannot do that. I pushed to the kaistore.

jpita:
> > > > @stephanebisson can you send the app to the kaios store to see if everything works well when there?
> > > 
> > > 
> > > Done
> > 
> > 
> > you deployed the jio one?
> 
> We cannot do that. I pushed to the kaistore.

I meant that the code you pushed was the jio one.
because the indian languages appear on top instead of the western ones on the nokia device

PR #239, opened 2020-06-01
T253018: only gallery images should be selectable

medied:
Phabricator Link: https://phabricator.wikimedia.org/T253018

### Problem Statement

If an image is not supposed to be shown in the gallery, it shouldn't be selectable. Otherwise it leads to a buggy behavior where a user can click on an image (from article) and then see a different image in the gallery view. 

### Solution

Check whether the image in the article section is part of the gallery images before adding it as a `visibleLink`. This includes images shown in `QuickFacts`

### DIFF
--- a/src/hooks/useArticleLinksNavigation.js
+++ b/src/hooks/useArticleLinksNavigation.js
@@ -21,7 +21,8 @@ export const useArticleLinksNavigation = (
   lang,
   contentRef,
   linkHandlers = {},
-  dependencies = []
+  dependencies = [],
+  galleryItems = []
 ) => {
   const i18n = useI18n()
   const [links, setLinks] = useState([])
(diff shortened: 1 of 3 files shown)


src/hooks/useArticleLinksNavigation.js
```diff
@@ -168,3 +172,14 @@ const findVisibleLinks = container => {
...
 }
+
+const isImageInGallery = (galleryItems, link) => {
+  const aElement = link.querySelector('a')
+  if (!aElement) {
+    return false
+  }
+
+  const href = aElement.getAttribute('href')
+  const fileName = href.split(':')[1]
```
hueitan, line 183: `makeLinkClickEvent` uses the same method here, what about reuse the function how it finds `fileName`?

medied: Yea why not, maybe something like below if we start from anchor (because anchor is not always guaranteed within `isImageInGallery`)

```
const getFileNameFromAnchorElement = (aElement) => {
  // file name example in href : /wiki/File:Holly_Christmas_card_from_NLI.jpg
  // split to match the api file name
  const href = aElement.getAttribute('href')
  return href.split(':')[1]
}
```

hueitan: sounds good 👍

PR #240, opened 2020-06-02
Update cypress to run on firefox

hueitan:
Phabricator Link: -

### Problem Statement

Our environment on the phone is firefox browser

### Solution

Update the cypress configuration and our app feature flag to be able to run on firefox

### Note

_I believe our app won't be able to work correctly when KaiOS updates the gecko engine to the latest, for example, if you run our app on chrome, the article page navigation doesn't work well due to the reason we are now developing it for the older firefox. The good thing is we may not need the `isProd` utils for setting request header._

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
           name: Build and Test
           command: |
             npm install
-            node ./node_modules/cross-env/src/bin/cross-env.js DISABLE_REQUEST_HEADER=1 npm run build
+            npm run build:ci
             npm run banana
             npm run start &
             npm test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
           name: Build and Test
           command: |
             npm install
-            npm run build
+            node ./node_modules/cross-env/src/bin/cross-env.js DISABLE_REQUEST_HEADER=1 npm run build
             npm run banana
             npm run start &
             npm test
--- a/cypress/integration/article-view-spec.js
+++ b/cypress/integration/article-view-spec.js
@@ -177,9 +177,7 @@ describe('Article view', () => {
   it('check text size remains after changing on preview', () => {
     goToCatArticle()
     articlePage.getArticleText().should('have.attr', 'style', 'font-size: 16px;')
-    articlePage.selectOptionFromActionsMenu('sections')
-    articleMenuPage.selectOptionFromSections('Taxonomy')
-    cy.enter()
+    cy.downArrow().enter()
     articlePage.decreaseTextSize()
     articlePage.decreaseTextSize()
     popupPage.getText().should('have.attr', 'style', 'font-size: 14px;')
--- a/cypress/integration/article-view-spec.js
+++ b/cypress/integration/article-view-spec.js
@@ -177,7 +177,9 @@ describe('Article view', () => {
   it('check text size remains after changing on preview', () => {
     goToCatArticle()
     articlePage.getArticleText().should('have.attr', 'style', 'font-size: 16px;')
-    cy.downArrow().downArrow().downArrow().enter()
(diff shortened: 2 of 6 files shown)


cypress/integration/article-view-spec.js
```diff
@@ -177,7 +177,9 @@ describe('Article view', () => {
   it('check text size remains after changing on preview', () => {
     goToCatArticle()
     articlePage.getArticleText().should('have.attr', 'style', 'font-size: 16px;')
-    cy.downArrow().downArrow().downArrow().enter()
+    articlePage.selectOptionFromActionsMenu('sections')
```
jpita, line 180: you found the issue? awesome!
I was confused because it was working fine on my local...

jpita:
how can we know the firefox version installed on the CI?

cypress/integration/article-view-spec.js
hueitan: I was confused too

hueitan: chrome works, but the user sees the page as blank, but able to select the preview.
but i'm not sure how it looks like in the terminal cypress

jpita:
> how can we know the firefox version installed on the CI?

I think we can add 
`navigator.appVersion` 
in a 
`console.log`
in a test

jpita:
> > how can we know the firefox version installed on the CI?
> 
> I think we can add
> `navigator.userAgent`
> in a
> `console.log`
> in a test
no need for console.log, cypress prints it out
image

cypress/integration/article-view-spec.js
hueitan: image

hmmm when checking the video, it opens the reference preview not article preview ...

hueitan: Ok, I have a slightly changed to the behaviour here, due to the window.screen reason, we don't know where it goes when press down 3 times, therefore press 1 time to get more predictable location either next page or next section.

.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
           name: Build and Test
           command: |
             npm install
-            npm run build
+            node ./node_modules/cross-env/src/bin/cross-env.js DISABLE_REQUEST_HEADER=1 npm run build
```
stephanebisson, line 12: Please encapsulate that as a command in package.json

src/utils/cachedFetch.js
```diff
@@ -15,7 +15,7 @@ export const cachedFetch = (url, transformFn, cache = true) => {
   const promise = new Promise((resolve, reject) => {
     xhr.responseType = 'json'
     xhr.open('GET', url)
-    if (isProd()) {
+    if (isProd() && !isRequestHeaderDisabled()) {
```
stephanebisson, line 18: Does it make sense to have both conditions here?

hueitan:
> I'm not really convinced of this change, but I'm also not against it. Have we seen situations where new Firefox behaves more similar to old Firefox than Chrome does?

That's what my initial thought, our app runs on the older firefox browser. we can't use the exact same environment as the device in the cypress test. According to what @jpita found, page flip works in firefox but not chrome (that I can confirm on chrome), can you confirm this again?

> Also, I don't have new Firefox locally, will the tests still work with Chrome?

Yes, the test still works in Chrome.

I also don't have any strong opinion on this, running both chrome/firefox can give us an overview of the test spec, which both are awesome!

src/utils/cachedFetch.js
hueitan: agree with having only one condition here.

hueitan:
@jpita ready to be reviewed, please test local and app carefully.

jpita:
@hueitan can u fix the conflicts when u have some time please?

hueitan:
Done @jpita

PR #241, opened 2020-06-02
T248171: add additional logic for device when checking links action

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T248171

### Problem Statement

Device unable to have "goto" section feature

### Solution

due to the different pathname, add additional logic for device

### Note

browser works correctly now, but not device, therefore this change is for device

### DIFF
--- a/src/hooks/useArticleLinksNavigation.js
+++ b/src/hooks/useArticleLinksNavigation.js
@@ -103,7 +103,7 @@ export const useArticleLinksNavigation = (
 
 const makeLinkClickEvent = link => {
   const title = link.getAttribute('title')
-  if (title && link.pathname !== '/') {
+  if (title && !['/', '/index.html'].includes(link.pathname)) {
     if (title.includes(':') && INTERWIKI_KEYS.includes(title.split(':')[0])) {
       return { type: 'external', href: link.href }
     }
--- a/cypress/integration/article-view-spec.js
+++ b/cypress/integration/article-view-spec.js
@@ -177,7 +177,7 @@ describe('Article view', () => {
   it('check text size remains after changing on preview', () => {
     goToCatArticle()
     articlePage.getArticleText().should('have.attr', 'style', 'font-size: 16px;')
-    cy.downArrow().downArrow().downArrow().enter()
+    cy.downArrow().enter()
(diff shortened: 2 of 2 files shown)


src/hooks/useArticleLinksNavigation.js
```diff
@@ -103,7 +103,7 @@ export const useArticleLinksNavigation = (
 
 const makeLinkClickEvent = link => {
   const title = link.getAttribute('title')
-  if (title && link.pathname !== '/') {
+  if (title && !['/', '/index.html'].includes(link.pathname)) {
```
stephanebisson, line 106: Could you explain what's gong on here, give examples oaf paths this code encounters and how it handles them?

hueitan: in the normal markup with having title attribute, we would like to trigger the article preview

```html
<a href="/wiki/Sean_Chiplock" title="Sean Chiplock">Sean Chiplock</a>
```

in the idea markup going to the specific anchor within the same article

```html
<a href="#Related_characters">See below</a>
```

but it's normal to see it has the title attribute, which conflicts with the first statement we have

```html
<a href="#Related_characters" title="C">See below</a>
```

therefore the need to the extra guard of the title attribute with the pathname

---

the perfect solution will be, passing all the anchor name to the hook and check against when having the hash in the link

```pseducode
hash => check all anchor, it matches then go to the anchor section
if doesn't match, continue the logic of starting with title
```

since we pass gallery items in this pr https://github.com/wikimedia/wikipedia-kaios/pull/239, are we fine also pass the sections list within the hook?

stephanebisson:
@jpita this is ready for testing

medied:
I tested on banana phone and seems to be working well 👍

>hash => check all anchor, it matches then go to the anchor section
if doesn't match, continue the logic of starting with title

Agree this might be a future enhancement opportunity though

hueitan:
> Agree this might be a future enhancement opportunity though

Yes, it looks good for now in the current scope. Let's make it the future enhancement, and probably we will see more different markup coming, then would be great timing to do this.

_Note I don't create a new ticket for this enhancement, things might change so we can create when we plan to do the enhancement_

PR #244, opened 2020-06-09
First use schema

stephanebisson:
Phabricator Link: -

### Problem Statement

The previous schema was wrong.

### Solution

Use the new schema to log first time use of the app and include app_id. Also refactor useTracking to make appInstallId reusable.


src/utils/consent.js
```diff
@@ -8,10 +8,13 @@ export const isConsentGranted = () => {
 
 export const grantConsent = () => {
   const version = appVersion()
-  sendEvent('KaiOSAppConsent', 20122102, getAppLanguage(), { version })
+  sendEvent('KaiOSAppFirstRun', 20158341, getAppLanguage(), {
+    app_version: version,
```
medied, line 12: From [the schema](https://meta.wikimedia.org/wiki/Schema:KaiOSAppFirstRun) I'm seeing that `app_id` comes first and `app_version` second, confirm the order doesn't matter?

stephanebisson: Order of the keys in a json doc doesn't matter

src/utils/generateRandomId.js
```diff
@@ -0,0 +1,9 @@
+export const generateRandomId = () => {
+  const rnds = new Uint16Array(5)
```
medied, line 2: Just curious, does `rnds` stand for 'randoms' or something?

src/hooks/useTracking.js
```diff
@@ -1,38 +1,21 @@
...
-import { appVersion, sendEvent, isInstrumentationEnabled } from 'utils'
+import {
+  appVersion, appInstallId, generateRandomId, sendEvent,
+  isInstrumentationEnabled
+} from 'utils'
 
 const SCHEMA_NAME = 'InukaPageView'
 const SCHEMA_REV = 19883738
-const USER_ID_KEY = 'INUKA-PV-U'
 const SESSION_ID_KEY = 'INUKA-PV-S'
```
hueitan, line 9: just checking the latest schema, the `user_id` said "Randomly generated user id, expiring after 90 days of inactivity.", but the `appInstalledId` never expires once user starts using the app.

is that fine we don't set the time checking same as the session?

stephanebisson: This is generally true regardless of the implementation since all that data is deleted on the server after 90 days.

src/utils/generateRandomId.js
stephanebisson: yeah, it probably stands for 'randoms'... I stole this code from https://github.com/wikimedia/mediawiki/blob/master/resources/src/mediawiki.user.js#L47

PR #246, opened 2020-06-10
(POC) add rtl direction for article content

hueitan:
do not merge, more coming, and this solution is not generic enough.


stephanebisson:
I know this is a draft but here's some early feedback for you to think about (you may already know this)
* As soon as we start supporting RTL we'll likely have mixed-directionality in the UI. This means dir should always be set at the root of the content node with `dir="ltr"` or `dir="rtl"`. Instead of looking at it as `rtl: true/false`, think of it as `dir: ltr/rtl`.
* Let's rely on the `dir` returned by the API as much as possible and not on our own list of RTL languages. Most of the endpoint will return the dir of the content language.

hueitan:
close and I will open another to use `dir` instead `rtl` state

PR #249, opened 2020-06-15
T255397 Offline status for consent page

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T255397

### Problem Statement

When we accept the terms (on first app boot after install) a call is sent to the logs. When offline, a user can still use some basic functionality on the app.

### Solution

Show the offline panel UI and hide the CSK button

### Note

_higher priority review task_

### DIFF
--- a/src/components/Consent.js
+++ b/src/components/Consent.js
@@ -1,31 +1,40 @@
 import { h } from 'preact'
-import { useI18n, useSoftkey } from 'hooks'
+import { useI18n, useSoftkey, useOnlineStatus } from 'hooks'
 import { grantConsent, goto } from 'utils'
+import { OfflinePanel } from 'components'
 
 export const Consent = () => {
   const i18n = useI18n()
+  const isOnline = useOnlineStatus()
 
   const onAgree = () => {
-    grantConsent()
-    goto.search()
+    if (isOnline) {
+      grantConsent()
+      goto.search()
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/components/Consent.js
```diff
@@ -1,31 +1,40 @@
...
 
   useSoftkey('ConsentMessage', {
-    center: i18n('softkey-consent-agree'),
+    center: isOnline ? i18n('softkey-consent-agree') : '',
     onKeyCenter: onAgree,
     left: i18n('softkey-consent-terms'),
     onKeyLeft: goto.termsOfUse,
     right: i18n('softkey-consent-policy'),
     onKeyRight: goto.privacyPolicy,
     backspace: () => { window.exit() }
```
medied, line 24: I missed this from a [previous PR](https://github.com/wikimedia/wikipedia-kaios/commit/c2dfdd34eef5d68800b97c39f980a5e79c98fc97#diff-e2c03eb00e4bd950fb680e50334c2846), but we do use `onKeyBackspace` everywhere else in the code so this `backspace` breaks the consistency.

hueitan: This is where user exit the app when user stuck on the consent page or don't want to agree with our term, so I think that's fine.

hueitan: Unless we want the user back to onboarding screen here

PR #251, opened 2020-06-16
T255397 Consent page - Clear softkey when offline

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T255397

### Problem Statement

Show the offline panel and don't show the softkey when offline

### Solution

clear the softkey

### Note

_we can have util or dispatch to clear all softkey, given that this is the only case, therefore these all the `isOnline` state check for each_

### DIFF
--- a/src/components/Consent.js
+++ b/src/components/Consent.js
@@ -18,9 +18,9 @@ export const Consent = () => {
     center: isOnline ? i18n('softkey-consent-agree') : '',
     onKeyCenter: onAgree,
     left: isOnline ? i18n('softkey-consent-terms') : '',
-    onKeyLeft: () => isOnline && goto.termsOfUse,
+    onKeyLeft: () => { isOnline && goto.termsOfUse() },
     right: isOnline ? i18n('softkey-consent-policy') : '',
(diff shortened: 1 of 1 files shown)


src/components/Consent.js
```diff
@@ -17,10 +17,10 @@ export const Consent = () => {
   useSoftkey('ConsentMessage', {
     center: isOnline ? i18n('softkey-consent-agree') : '',
     onKeyCenter: onAgree,
-    left: i18n('softkey-consent-terms'),
-    onKeyLeft: goto.termsOfUse,
-    right: i18n('softkey-consent-policy'),
-    onKeyRight: goto.privacyPolicy,
+    left: isOnline ? i18n('softkey-consent-terms') : '',
```
stephanebisson, line 20: Since almost all the keys are conditional, you could consider having 2 different objects: one fully configured and one with a bunch of blank keys and decide which one to pass to `useSoftkey` based on `isOnline`.

hueitan: Thanks! You remind me how we handle this is onboarding page with the replace params.

medied: Nice

PR #252, opened 2020-06-17
T236317: add specific arrow keys handlers

medied:
Phabricator Link: https://phabricator.wikimedia.org/T236317

### Problem Statement

This is a follow up to #216, it was requested to update `Feedback` navigation to:

> Use up/down keys on D-PAD to switch control between the text box and fine print and use left/right keys to switch between links

### Solution

I [shared my takeaways](https://phabricator.wikimedia.org/T236317#6230469) in the Phab ticket, it became evident to me that having both up/down and right/left arrow key handlers for a _single_ screen is basically not common/present anywhere else in the app, and therefore slightly against overall app consistency. If we want to go through with it though, to achieve the requested behavior then I'm proposing a `onKeyArrow*` for each of the four options and using `setNavigation` to select. Feels like some of the logic could be abstracted out to simplify the code but that may be part of a bigger refactor. 

Note that deleting characters of `message` is slightly buggy on the browser (if you try to delete a character that's not the last character), but I've tested and confirmed it works as expected on device (banana phone)

### DIFF
--- a/src/components/Feedback.js
+++ b/src/components/Feedback.js
@@ -89,11 +89,6 @@ export const Feedback = ({ close }) => {
     }
   }
 
-  const handleUpOrDownArrowKey = () => {
-    const { index } = getCurrent()
-    index === 0 ? setNavigation(1) : setNavigation(0)
-  }
-
   const getTextareaElement = () => containerRef.current.querySelector('textarea')
 
   const blurTextarea = () => {
@@ -109,9 +104,7 @@ export const Feedback = ({ close }) => {
     onKeyBackspace,
     onKeyCenter,
     onKeyArrowRight,
-    onKeyArrowLeft,
-    onKeyArrowDown: handleUpOrDownArrowKey,
-    onKeyArrowUp: handleUpOrDownArrowKey
+    onKeyArrowLeft
   }, [message, isOnline])
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


medied:
Also, when working with this I found a bug with the backspace behavior: it doesn’t show the cancel confirmation popup if you’re focused on a link and hit backspace. The following would work better:

```
const onKeyBackspace = () => {
    if (message && getCurrent().type !== 'TEXTAREA') {
        showCancelConfirmation()
    } else if (message) {
        setMessage(message.slice(0, -1))
    } else {
        close()
    }
}
```

Okay to commit this fix in this PR if we end up merging? We should fix that bug either way

hueitan:
> Okay to commit this fix in this PR if we end up merging? We should fix that bug either way

That's fine to have the fix in this PR.

> Note that deleting characters of message is slightly buggy on the browser (if you try to delete a character that's not the last character), but I've tested and confirmed it works as expected on device (banana phone)

Regarding this, best to have the fix (same to search screen, language page), it works on device but codewise looks incorrect.

medied:
> What I suggested here should fix it #216 (comment)

Yea! Actually forgot to mention that, in the debugging session I had yesterday I did come across your comment and tried it out but didn't see any change. It works on your machine?

medied:
@hueitan yea that does simplify the code in my opinion, check out the last commit.`useSoftkey`  gets a bit harder to read but I think the trade-off is worth it. What do you think? This fixes the bug I had found with `onKeyBackspace` too. Should I go ahead and update the `onKeyBackspace` for `Language` and `Search` in this same PR?

hueitan:
Looks good, yeah let's fix those in this branch.

jpita:
didn't notice this was not in QA yet, should I revert the commit?

medied:
Yea please revert @jpita, there are a couple commits still left in this PR

PR #255, opened 2020-06-22
Include the circleci build number in the app version

stephanebisson:
Phabricator Link: -

### Problem Statement

I don't know which version of the app is published to the store.

### Solution

Include the CircleCI build number as the 4th digit of the version.

### DIFF
--- a/scripts/manifest.js
+++ b/scripts/manifest.js
@@ -159,8 +159,9 @@ manifest.display = translations['en'].name
 manifest.subtitle = translations['en'].subtitle
 manifest.description = translations['en'].description
 
-// Use the package version as the source of truth
-manifest.version = require('../package.json').version
+// Use the package version and build number
+const buildNumber = process.env.CIRCLE_BUILD_NUM || 0
+manifest.version = require('../package.json').version + '.' + buildNumber
 
 // Send output to stdout
 console.log(JSON.stringify(manifest, null, 2))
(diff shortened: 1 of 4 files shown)


medied:
> Include the CircleCI build number as the 4th digit of the version.

Do we need this to be true everywhere `appVersion` is recorded? (consent.js, sendFeedback.js, etc) Or is 3 digit version enough in other cases?

stephanebisson:
> > Include the CircleCI build number as the 4th digit of the version.
> 
> Do we need this to be true everywhere `appVersion` is recorded? (consent.js, sendFeedback.js, etc) Or is 3 digit version enough in other cases?

It would probably be best to have the same version captured everywhere. Maybe I can just encapsulate that into `appVersion()` so it is used everywhere.

hueitan:
Ideally, once we release our first version `1.0.0`, every release after will need to increment the version digit.

Do you sitll need the build number if we strictly follow the version rule?

stephanebisson:
> Ideally, once we release our first version `1.0.0`, every release after will need to increment the version digit.
> 
> Do you still need the build number if we strictly follow the version rule?

Relying on the package version alone is awkward because we can't go back and pick a build to release. We have to decide to release, change the version, wait for it to build and pick this one. After that, we don't know if we released the first or any other build with that version number. For example, we can't pick the build just before a big risky feature was merged.

Using the CI build number guarantees that each build is unique and identifiable. 

We should still use a versioning scheme to identify work phases (minor, major).

PR #264, opened 2020-07-28
Language lists use buttons

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T254311

### Problem Statement
Language lists use radio buttons. To select a language, the user has to select the radio button first and then click the left softkey to go back.

### Solution
Use buttons so that selecting a language applies it immediately. This is also how KaiOS does it.

### Note
Adding a npm task to run the browser tests with Chrome. Firefox doesn't work for me.

### DIFF
--- a/cypress/integration/article-view-spec.js
+++ b/cypress/integration/article-view-spec.js
@@ -25,7 +25,6 @@ describe('Article view', () => {
     cy.get('input').type('português')
     cy.get('.description').should('have.text', 'Gato')
     cy.downArrow().enter()
-    cy.clickDoneButton()
     articlePage.title().should('have.text', 'Gato')
   })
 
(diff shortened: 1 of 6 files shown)


hueitan:
You may work on it right now, the styling for `listview` and `radiolistview` has slightly different in the font size

| ListView | RadioListView |
| --- | --- |
| Screen Shot 2020-07-28 at 22 32 46 | Screen Shot 2020-07-28 at 22 32 33 |

stephanebisson:
> You may work on it right now, the styling for `listview` and `radiolistview` has slightly different in the font size

Good observation but I don't know if it should be changed here. I've looked around and it looks like all lists have 17px font except those with 2 lines (title, desc) that are 15px. There is some logic there. I'll bring it up with desigbn.

hueitan:
NOTE : One thing missing here is the support of RTL list item in the article language list, but no worry, I'm going to add them in the UI language PR.

PR #265, opened 2020-07-31
POC - RTL Support for UI Language

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T252827

### Problem Statement

We have RTL Support for Content Language now, then it's time for the RTL Support for UI Language

### Solution

In Content language, we need to pass the content language direction state into each component, in the UI language part, instead of passing the ui language direction state into all of the components, we set the direction to the body tag element, most of the cases CSS can detect it and do all the styling correctly.

### Note

more coming

 - [ ] General solution of how we handle left/right (hardware/arrow) key in CSS style and action handler

### DIFF
--- a/cypress/page-objects/article-menu-page.js
+++ b/cypress/page-objects/article-menu-page.js
@@ -1,6 +1,6 @@
 export class ArticleMenuPage {
   getPreviousArticleName () {
-    return cy.get('div.info > div.description')
+    return cy.get('.info > .description')
   }
 
   selectOptionFromArticleMenu (option) {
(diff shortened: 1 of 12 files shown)


hueitan:
_This is ready for code review, screenshot comparison is coming late._

hueitan:
> instead of passing the ui language direction state into all of the components, we set the direction to the body tag element, most of the cases CSS can detect it and do all the styling correctly.

Due to the ui language direction state change need for softkey to flip the key, we need to set state to change the direction. Turns out, I may need to reconsider the internal state opportunity. Therefore closing this and opening another one when softkey is handled in the smarter way.

PR #272, opened 2020-09-15
Fix outline fragmentation by using border

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T260391

### Problem Statement

References have an outline around every single character. See task for screenshots.

### Solution

bidi-isolate the whole reference link so it doesn't switch direction and has a single outline.

### DIFF
--- a/style/article.less
+++ b/style/article.less
@@ -172,8 +172,7 @@
 
 					@{data-selected-selector} {
 						outline: solid 1px rgba( 51, 102, 204, 0.5 );
-						background-color: #ffeaf3;
-						display: inline-block;
+						background-color: #eaf3ff;
 					}
 				}
 
--- a/style/article.less
+++ b/style/article.less
@@ -172,17 +172,14 @@
 
 					@{data-selected-selector} {
 						outline: solid 1px rgba( 51, 102, 204, 0.5 );
-						background-color: #eaf3ff;
+						background-color: #ffeaf3;
+						display: inline-block;
 					}
 				}
 
 				span.mw-ref {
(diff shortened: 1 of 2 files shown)


hueitan (changes requested):
Still missing in this example article `C` in hebrew
Screen Shot 2020-09-16 at 14 00 37

---

a better solution is to apply the `unicode-bidi: bidi-override` here to cover all cases.
https://github.com/wikimedia/wikipedia-kaios/blob/ec3051f1fbdf2562850c1a0e011fcb41586fcf67/style/links.less#L1-L5

stephanebisson:
> Still missing in this example article `C` in hebrew

Good catch! I was looking at references but the issue can happen with any selectable text.

> a better solution is to apply the `unicode-bidi: bidi-override` here to cover all cases.

It was wrong to use bidi-override in the first place. It breaks the layout in cases where the browser was correctly guessing. 

I found outline to be simply broken in Firefox under these conditions. Here's a minimal test case: https://jsbin.com/konitelaya/edit?html,css,output

hueitan (approved):
It's a good option to replace the buggy outline!

jpita:
this has broken the image selection, when selected the image moves, sometimes to a different page
Large GIF (248x396)

hueitan (changes requested):
Good x2 catch @jpita, I can confirm the problem.

_Great that we haven't merged the PR._

stephanebisson:
@jpita should be fixed. Ready for testing again.

PR #273, opened 2020-09-15
Limit length of search input according to search API restriction

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T262591

### Problem Statement

The search input allows an unlimited number of characters but the API allows only 255.

### Solution

Limit search input to 255 characters to avoid useless and noisy error.


hueitan (approved):
Best quick solution!!

PR #274, opened 2020-09-16
T262602 article history

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T262602

### Problem Statement

Article history can become large and fill up local storage.

### Solution

There's no reason to put it in local storage. Store in a regular variable and limit its size to 100.

### DIFF
--- a/src/utils/articleHistory.js
+++ b/src/utils/articleHistory.js
@@ -1,10 +1,14 @@
 import { normalizeTitle } from 'utils'
 
+const MAX = 100
 const list = []
 
 const add = (lang, title) => {
   const normalizedTitle = normalizeTitle(title)
   list.push({ lang, title: normalizedTitle })
+  if (list.length > MAX) {
+    list.shift()
+  }
 }
 
 const prev = () => {
--- a/src/utils/articleHistory.js
+++ b/src/utils/articleHistory.js
@@ -1,25 +1,14 @@
 import { normalizeTitle } from 'utils'
 
-const KEY = 'article-history'
-
-const get = () => {
-  return JSON.parse(localStorage.getItem(KEY))
-}
-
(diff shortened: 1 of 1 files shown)


medied (approved):
I like this simplification

hueitan (approved):
agree with that, i don't see the reason why using localstorage for the article history

jpita:
This branch broke the ability to open references and it wasn't supposed to touch any such thing apart from the history.
Can we investigate what happened and check if this was a directly related issue or a "let me fix this while I'm here" issue?

jpita:
also once it's fixed remind me to add a test to this before merging

stephanebisson:
> This branch broke the ability to open references and it wasn't supposed to touch any such thing apart from the history.
> Can we investigate what happened and check if this was a directly related issue or a "let me fix this while I'm here" issue?

Could you be more specific about what you see being broken?

hueitan:
well, the reference preview is broken on master as welll ...h mmm

jpita:
> > This branch broke the ability to open references and it wasn't supposed to touch any such thing apart from the history.
> > Can we investigate what happened and check if this was a directly related issue or a "let me fix this while I'm here" issue?
> 
> Could you be more specific about what you see being broken?

sorry, was late yesterday and forgot about this
I added a bit more details here https://phabricator.wikimedia.org/T262602#6468465
basically once you click a reference the app goes black as it would show the preview popup but doesn't show the popup.
then we need to close the app cause no other action is available.

I tried on the released app and doesn't happen, didn't try on master.

Thanks for creating the ticket @hueitan

hueitan:
@jpita you can re-qa this pr again, as the reference preview issue has been fixed in another ticket pr.

PR #271, opened 2020-09-10
Update most packages to fix npm audit warnings

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T253141

### Problem Statement

Some packages that were used have known vulnerabilities.

### Solution

Update the packages to versions that have been fixed.

### Note

The remaining issues can be traced back to `b2g-cli-tool`. This is a very old tool that we used to send the KaiOS app to the test devices. It was build for FirefoxOS and it's unlikely to be updated ever again. This is NOT used in production or even CI, only by developers to test the app.

PR #275, opened 2020-09-17
Fix reference preview by downgrading banana-i18n

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T263109

### Problem Statement

banana-i18n was update from 1.2.1 to 1.3.1 (even though it was not needed) in #271 to fix npm audit warnings. That version turns out to error out on the reference dialog title message. Will follow up with banana-i18n separately. 

### Solution

Downgrade banana-i18n from 1.3.1 to 1.2.1

### DIFF
--- a/cypress/integration/references-popup.js
+++ b/cypress/integration/references-popup.js
@@ -9,7 +9,7 @@ describe('references', () => {
     cy.navigateToHomePage()
   })
 
-  it('check preview for references opens', () => {
+  it.skip('check preview for references opens', () => {
     goToCatArticle()
     cy.downArrow()
     cy.rightArrow()
--- a/cypress/integration/references-popup.js
+++ b/cypress/integration/references-popup.js
@@ -1,0 +1,24 @@
+/// <reference types="Cypress" />
+
+import { ArticlePage } from '../page-objects/article-page'
+
+const articlePage = new ArticlePage()
+
+describe('references', () => {
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


hueitan (approved):
Confirm the fix.

jpita:
How can we avoid this in the future (besides having more automated tests for all the user flows possible)?

jpita:
can anyone run `npm test` on their machine to see the result of the tests?
@stephanebisson the version of firefox on circleci is 66 while the latest version is 80, is there any way to update circleci?

jpita:
the test passes on my local without the `--silent` flag but fails with the `--silent` flag with firefox 80.
on circle ci fails even without the flag on firefox 66.
I created an issue on the cypress repo so I'll put this test on skip https://github.com/cypress-io/cypress/issues/8610

cypress/integration/references-popup.js
```diff
@@ -0,0 +1,24 @@
...
+  beforeEach(() => {
+    cy.navigateToHomePage()
+  })
+
+  it.skip('check preview for references opens', () => {
+    goToCatArticle()
+    cy.downArrow()
+    cy.rightArrow()
+    cy.enter()
+    cy.get('.ref-title').should('be.visible').should('contain.text', 'Reference [10]')
```
hueitan, line 17: Hey @jpita, mine was opening one article preview in my local

image

PR #276, opened 2020-09-21
T260114 Configure input element mic icon

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T260114

### Problem Statement

input mic icon overlap with the placeholder in RTL language

### Solution

update the input css

### Note

**Tested on KaiOS NOKIA flip phone**; if you don't see any mic icon on your device, then this change doesn't affect it.

| LTR | RTL (1) | RTL (2) | RTL (3) |
| -- | -- | -- | -- |
| image | image | image | image |

### DIFF
--- a/style/input.less
+++ b/style/input.less
@@ -1,0 +1,10 @@
+* {
+	[ dir='rtl' ] {
+		input {
+			&[ nav-selected='true' ] {
+				// mic icon position
+				background-position: left 6.8px center !important;
+			}
+		}
+	}
+}
--- a/style/input.less
+++ b/style/input.less
@@ -1,10 +1,6 @@
-* {
-	[ dir='rtl' ] {
-		input {
-			&[ nav-selected='true' ] {
-				// mic icon position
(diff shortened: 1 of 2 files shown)


style/input.less
```diff
@@ -0,0 +1,10 @@
+* {
```
stephanebisson, line 1: I think it would be fine to write this a bit more compact since the intermediate levels don't have any style.

```
*[dir=rtl] {
  input[nav-selected=true] {
    // style here
  }
}
```

style/input.less
```diff
@@ -0,0 +1,6 @@
+*[ dir='rtl' ] {
+	input[ nav-selected='true' ] {
```
stephanebisson, line 2: I'm confused by the `nav-selected` here... The position of the icon changes when it becomes selected?

hueitan: Yes, the icon shown only when the input is selected.

stephanebisson (approved):
I don't have the right device to test this... let's give it a try.

PR #277, opened 2020-10-07
remove unwanted package dev dependencies

hueitan:
I don't see a reason it stays in dev dependencies, that's the dependencies of the dev dependencies, and we don't use it in our source code.


hueitan:
oh, is that because of the result of `npm audit`?

stephanebisson:
> oh, is that because of the result of `npm audit`?

Yes, this is to up those transient dependencies.

PR #278, opened 2020-10-08
Upgrade banana-i18n version to 2.0.0

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T262596

I can't reproduce the error, and this is what I found and possibly to prevent the error

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2623,9 +2623,9 @@
       "dev": true
     },
     "banana-i18n": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/banana-i18n/-/banana-i18n-1.3.1.tgz",
-      "integrity": "sha512-9Uo441d1ahU4KQntyAh11gxjDpNqUVd2X32v1SYSnIf75lOAB2hCvaWYzuuiB1q7qQabr1wR7rgFUPwGq4r5oA=="
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/banana-i18n/-/banana-i18n-2.0.0.tgz",
+      "integrity": "sha512-dwnkYF1A/U36FO7UKw35m98Qtd61BWgrF0jdsyy6QviK6YWy+S9cY2r//v6K9oUnkqT++ChFCm0eof7fIoDYbw=="
     },
     "base": {
       "version": "0.11.2",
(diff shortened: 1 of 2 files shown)


stephanebisson (changes requested):
Banana v1.3.0 is buggy. We can't open the reference popup.

hueitan:
> Banana v1.3.0 is buggy. We can't open the reference popup.

hmmm let me look into that

hueitan:
I have opened the issue from `banana-i18n` https://github.com/wikimedia/banana-i18n/issues/40, does it worth if we change all to have the form in `\\[\\]`?

stephanebisson:
> I have opened the issue from `banana-i18n` [wikimedia/banana-i18n#40](https://github.com/wikimedia/banana-i18n/issues/40), does it worth if we change all to have the form in `\\[\\]`?

There's already an issue where this is discussed: https://github.com/wikimedia/banana-i18n/issues/37

Santhosh said he would produce a new release where special parsing of `[]` can be disabled (or is disabled by default). We will upgrade to that version when it is released.

Let's not do `\\[\\]` in the message. We could do a string replace on the message before passing it to `i18n` temporarily as a quick fix but we would need a comment and maybe a task to make sure we go back to fixing it properly.

hueitan:
> > I have opened the issue from `banana-i18n` [wikimedia/banana-i18n#40](https://github.com/wikimedia/banana-i18n/issues/40), does it worth if we change all to have the form in `\\[\\]`?
> 
> There's already an issue where this is discussed: [wikimedia/banana-i18n#37](https://github.com/wikimedia/banana-i18n/issues/37)
> 
> Santhosh said he would produce a new release where special parsing of `[]` can be disabled (or is disabled by default). We will upgrade to that version when it is released.

Thanks, I just know there are issue and progress on this! 

> We could do a string replace on the message before passing it to `i18n` temporarily as a quick fix but we would need a comment and maybe a task to make sure we go back to fix it properly.

Let's not doing this, going through all the messages before passing to banana is too expensive, i would rather wait for the fix.

stephanebisson:
banana-i18n v2.0.0 is out. It has parsing of `[]` disabled by default and should fix the production error from the linked task. We can update this PR to use it.

PR #279, opened 2020-10-08
T262592: fix production error ('v/h is undefined')

medied:
Phabricator Link: https://phabricator.wikimedia.org/T262592

### Problem Statement

Our Kibana dashboard shows the following error info

```
message: v is undefined

---

stack trace:

getInitialLangList/<@app://kaios.wikipedia.org/dist/main.js:1:395829
getInitialLangList@app://kaios.wikipedia.org/dist/main.js:1:395809
useSearchLanguage/<@app://kaios.wikipedia.org/dist/main.js:1:395323
q@app://kaios.wikipedia.org/dist/main.js:1:283028
I.__c/</h.__h<@app://kaios.wikipedia.org/dist/main.js:1:285570
I.__c/<@app://kaios.wikipedia.org/dist/main.js:1:285531
I.__c@app://kaios.wikipedia.org/dist/main.js:1:285471
T@app://kaios.wikipedia.org/dist/main.js:1:267101
_/<@app://kaios.wikipedia.org/dist/main.js:1:262636
_@app://kaios.wikipedia.org/dist/main.js:1:262455

---

App version: 1.0.0.1231

```

### Solution

As part of the initial investigation, I noticed in the Kibana dashboard that for every `v is undefined` error there is a corresponding `h is undefined` error. They come in pairs. Each pair shares the `@timestamp` and `http.client_ip` address, so this led me to believe [T262594](https://phabricator.wikimedia.org/T262594) may be related to this [T262592](https://phabricator.wikimedia.org/T262592). One error points to the `RadioListView` component and the other to the `getInitialLangList` method, both used when changing app language (`getInitialLangList` sets the items used by the `RadioListView` when there's no query).

The above led us to following hypothesis: the `lang` variable passed to `getInitialLangList` (from `getAppLanguage`) may be a language that's not included in the long `languages` list in `utils/languages.js`. If `items` get 'unshifted' [1] with an undefined or some faulty value in `getInitialLangList`, that would likely cause issues in `RadioListView`. To test that hypothesis, I tried to:

> ...set your device to a language that doesn't exist in the app to check if it replicates some of those issues. One way to do that is to set the app language to `es`, then comment out `es` in the code and reload the app.

And as a result, saw the following issue appear which is consistent with what T262594 and T262592 hint at:

image

As a solution I'm proposing to only unshift the list if the target language is found. 

Are there any more tests we can think of? 

---

### Note

[1]

```
Welcome to Node.js v13.8.0.
Type ".help" for more information.
>
> arr = [1, 2]
[ 1, 2 ]
>
> arr.unshift(0)
3
>
> arr
[ 0, 1, 2 ]
>
> arr.unshift(undefined)
4
>
> arr
[ undefined, 0, 1, 2 ]
>
> arr.unshift('undefined')
5
>
> arr
[ 'undefined', undefined, 0, 1, 2 ]
> undefined
undefined
```

### DIFF
--- a/src/hooks/useSearchLanguage.js
+++ b/src/hooks/useSearchLanguage.js
@@ -42,7 +42,10 @@ const getInitialLangList = lang => {
   const list = prioritizedLanguages
 
   if (!list.find(language => language.lang === lang)) {
-    list.unshift(allLanguages.find(language => language.lang === lang))
+    const targetLanguage = allLanguages.find(language => language.lang === lang)
+    if (targetLanguage) {
+      list.unshift(targetLanguage)
+    }
   }
 
   return list
--- a/src/utils/languages.js
+++ b/src/utils/languages.js
@@ -1509,6 +1509,9 @@ export const getDirection = langCode => {
 }
 
 export const isSupportedForReading = langCode => {
+  if (!languages.find(language => language.code === langCode)) {
+    return false
+  }
   return missingFont.indexOf(langCode) === -1
 }
 


hueitan:
Would put a guard during the initialization of setting device language `getCurrentDeviceLanguage` enough for the entire application? so we don't have to worry about the unknown language in any section later.

medied:
Hmm interesting idea, maybe as an extra validation besides the `getInitialLangList` update? Trying to think what would be the best way to structure that extra check though

What does the `missingFont` list represent exactly? I see it gets checked through `getCurrentDeviceLanguage` --> `isSupportedForReading`

hueitan:
> Hmm interesting idea, maybe as an extra validation besides the `getInitialLangList` update? Trying to think what would be the best way to structure that extra check though
> 
> What does the `missingFont` list represent exactly? I see it gets checked through `getCurrentDeviceLanguage` --> `isSupportedForReading`

https://github.com/wikimedia/wikipedia-kaios/blob/84a6aa0b4bd87f3cd7b396dd213c6db1d17190d4/src/utils/languages.js#L1588-L1594

the `navigatorLang` can be the lang that is not in allLanguages or missingFont, i thinkthe better place to have is in `isSupportedForReading`

1. if the lang isn't in the all languages list, then not supported;
2. if it is, check if it is supported for reading.

medied:
Yea I think you're right, it's reasonable to include that extra check in `isSupportedForReading`

hueitan (approved):
Looks good.

PR #282, opened 2020-11-17
T267289: Update highlighter styling

medied:
Phabricator Link: https://phabricator.wikimedia.org/T267289

### Problem Statement

Current search results highlighter is too light to be noticed

### Solution

Update styling as outlined by design

### DIFF
--- a/src/utils/languages.js
+++ b/src/utils/languages.js
@@ -5,7 +5,7 @@ const languages = [
   {
     canonical_name: 'English',
     code: 'en',
-    name: 'English (Default)'
+    name: 'English'
   },
   {
     canonical_name: 'Dutch',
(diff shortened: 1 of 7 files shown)


hueitan (approved):
Confirm looks good!

stephanebisson (commented):
Since all lists are based on `Listview` and `RadioListview`, could we update the style for those and remove the overrides in all other places?

medied:
> Since all lists are based on Listview and RadioListview, could we update the style for those and remove the overrides in all other places?

Looks like you found a good way to make the code more DRY. Good catch!

stephanebisson (approved):

stephanebisson (approved):

PR #283, opened 2020-11-17
T262249 add confirm dialog before exit the app in search page

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T262249

### Problem Statement

When a user uses the back key repeatedly, the app quits, but this behaviour is unexpected.

### Solution

Use Confirm Dialog component before quit

_Note : When user has search input box focus then press back key, there is a need to deal with the focus element_

### Note

Screenshot
image


medied (commented):
Code looks good, I left a comment in the Phabricator thread regarding a small language feedback though: https://phabricator.wikimedia.org/T262249#6637184

stephanebisson (changes requested):
Looks good.

I found an issue. I don't know if it's a deal breaker but I would like you to look into it.

If you hit back many times quickly the app becomes confused. The RSK changes from "Exit" to "Settings" but the popup stays open.

hueitan:
> If you hit back many times quickly the app becomes confused. The RSK changes from "Exit" to "Settings" but the popup stays open.

Another step to reproduce on the desktop browser, on the search screen, hold the `backspace` button for few seconds, then the softkey is mess up.

hueitan:
> I found an issue. I don't know if it's a deal breaker but I would like you to look into it.

This is fixed now, great you notice this behaviour!

stephanebisson (approved):

src/components/ConfirmDialog.js
```diff
@@ -10,7 +10,7 @@ export const ConfirmDialog = ({
 
   const onDiscardFn = () => {
     onDiscard()
-    close()
```
hueitan: this has side-effect, it should only close one level from the dialog! will look close into it!

hueitan: so in general the bug comes from the popup `close` and `show` function, the state doesn't fully update yet but the show/close event is called, so the softkey popup state is mess up with it. the fixed should be considered in https://phabricator.wikimedia.org/T269158

PR #286, opened 2020-11-30
T244548 Set up preview counting for KaiOS app

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T244548

### Solution

Similar to `useTracking` code, reuse the `sentEvent` util for the data sending

### Note

Schema url : https://meta.wikimedia.org/wiki/Schema:VirtualPageView


src/hooks/useArticlePreviewTracking.js
```diff
@@ -0,0 +1,42 @@
...
+) => {
+  if (!isInstrumentationEnabled()) {
+    return
+  }
+
+  if (page === undefined || source === undefined) {
+    return
+  }
+
+  const logInukaPageView = () => {
```
neilpquinn, line 18: We should probably update the name of this function :)

hueitan: Yes omg !!! Thanks @neilpquinn

src/hooks/useArticle.js
```diff
@@ -3,7 +3,14 @@ import { useI18n } from 'hooks'
 import { getArticle, getArticleMediaList, getSuggestedArticles } from 'api'
 import { canonicalizeTitle } from 'utils'
 
+let currentArticle
```
stephanebisson, line 6: Adding global state like that is not great. It took me a while to figure out the whole logic here. It's clever, and it works, but I would like us to explore more conventional (or "boring") options.

One such option is to give ArticlePreview the required info about the source on open
```
showArticlePreview({ title, lang, dir, <info about the current article> })
```

src/hooks/useArticlePreviewTracking.js
```diff
@@ -0,0 +1,42 @@
...
+const SCHEMA_REV = 19883675
+
+export const useArticlePreviewTracking = (
+  page, source, language
+) => {
+  if (!isInstrumentationEnabled()) {
+    return
+  }
+
+  if (page === undefined || source === undefined) {
```
stephanebisson, line 14: When is this called without `page` or `source`?

stephanebisson (changes requested):

src/hooks/useArticle.js
hueitan: I start with the conventional options, but then it's really bad =( but anyway, let me try again with it.

src/hooks/useArticlePreviewTracking.js
hueitan: _Just in case `useArticle()` is being called before the `useArticle( x, y )`, to make sure it returns empty {} instead of error._

src/hooks/useArticle.js
hueitan: indeed, yes, `useArticleLinksNavigation` has `galleryItems` that also come from the source, we can have everything as the `source`, so it can pass to `showArticlePreview`

hueitan:
Ready for review again @stephanebisson

stephanebisson (commented):
It's annoying to propagate a bunch of article info through many layers but it makes the code very simple and predictable.

Will test tomorrow.

src/hooks/useArticlePreviewTracking.js
```diff
@@ -0,0 +1,42 @@
...
+      /* eslint-enable camelcase */
+    }
+
+    sendEvent(SCHEMA_NAME, SCHEMA_REV, language, event)
+  }
+
+  useLayoutEffect(() => {
+    if (!page) {
+      return
+    }
```
hueitan, line 34: so this is still needed here, the summary is a state and start with null until the api request is done

hueitan:
> It's annoying to propagate a bunch of article info through many layers but it makes the code very simple and predictable.

Agree, I see the benefit of this solution now! it's easier to track and predict all the properties.

src/hooks/useArticlePreviewTracking.js
stephanebisson: Why `useLayoutEffect` specifically instead of `useEffect`?

Layout effect is blocking and should only be used to prevent flickering on visual effects.

PR #289, opened 2020-12-02
T269158 Confirm dialog - Close event

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T269158

### Problem Statement

The confirm dialog close event has `close` `closeAll` option, but it doesn't allow us to choose which one

### Solution

Make the confirm dialog utils/component more general
 - [x] 7a714c2 Reverse the previous changes of the key, now LSK `close()` and RSK `closeAll()` as the setting for confirm dialog utils
 - [x] 9a51818 306e564 prevent duplication of the stack id
 - [x] not allowing `close( <num_of_popup_closed> )`, << we don't have this cases yet, therefore ignore it.
 - [x] 5b4937369065bd61e78c3b8be8f428bfa8b7966b Update Feedback page to use `ConfirmDialog`
 - [x] dc8619d211477e3e9fa425e38a648bb89f46a06c Add test spec to prevent the test fail again in the future

### Note

touch to the confirm dialog and popup related code, including hook

### DIFF
--- a/cypress/integration/special-cases.js
+++ b/cypress/integration/special-cases.js
@@ -70,5 +70,6 @@ describe('special cases tests', () => {
     cy.get('.info').should('have.text', 'Go to Section "Related characters"')
     cy.backspace()
     cy.get('.info').should('not.be.visible')
+    cy.get('div.quickfacts table.infobox').should('be.visible')
   })
 })
--- a/src/components/ConfirmDialog.js
+++ b/src/components/ConfirmDialog.js
@@ -10,7 +10,7 @@ export const ConfirmDialog = ({
 
   const onDiscardFn = () => {
     onDiscard()
-    closeAll()
+    close()
   }
 
   useSoftkey('ConfirmDialog', {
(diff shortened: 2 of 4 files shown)


hueitan:
This is the detail change of the commit 9a518187f3050c2eaef078fe83440bafff81c645

In the current implementation, **duplication of the popup dialog** is allowed but we don't have the case, if we do so it may bring the infinite loop from the popup, for example, A has B then B has A.

I suggest **avoiding looping** (the repeated component popup stack in the list) by using id.

| Before | After |
| --- | --- |
| image | image |

> show - event called from usePopup
> close - event called from usePopup
> (array value) - state id

**Before**, `0.7135227832276235` `0.10885823320988941` are both the same component, but due to the event calling (state setting is async in hook) it has 2 same component in the stack

**After**, because it uses the component name as the id, so any repeated id will be prevented.

---

This is a rare edge case, hope the explanation is easy to understand 😅

hueitan:
Ready for review, I expect some feedback and comment as it changes the internal works for popup @medied @stephanebisson

PR #290, opened 2020-12-07
T267293: Update font size change option

medied:

### Problem Statement

We learned after some research that the feature to change the font size is confusing for some users.

### Solution

Update the font size change mechanism to the new proposed design.


stephanebisson (commented):
The way you've modified the feature makes sense but it may looks overly complex to someone who doesn't know how it used to work.

If you take a step back, is it how you would build it from the ground up or can you think of a simpler solution?

src/hooks/useArticleTextSize.js
```diff
@@ -2,12 +2,14 @@ import { useState, useLayoutEffect } from 'preact/hooks'
 import { useSoftkey } from 'hooks'
 import { articleTextSize } from 'utils'
 
-export const useArticleTextSize = (origin, dependencies = []) => {
```
hueitan, line 5: I really don't like the idea of having `withSoftkeys` param here, it used to be apply this softkey everywhere but it's now allowed in `TextSize` screen only.

**Quick idea :**

1) the `useArticleTextSize` in most cases are doing only `articleTextSize.init()` only
2) move the `useSoftkey` only in `TextSize` component?

**Another broader idea :**

1) Since using `origin` a lot in every place, and this string becomes important how we maintain the softkey, let's maintain the `origin` as the object which holds all the origin string value. {}
2) when having this object, within`useArticleTextSize` we only enable the softkey in `origin.textSize` << means textSize screen
3) OR combined the idea above.

medied: Yea actually looking back at this idea, I dislike it too, not sure what I was thinking. How about https://github.com/wikimedia/wikipedia-kaios/pull/290/commits/6f9597da50c0af5765b8b564572e9a87dbbee14f? Indeed we can simplify a bit by having the hook only `init` components that wish to be 'subscribed' to the font size changes. And then abstract the actual softkey setting part only to `TextSize` component.

medied:
> The way you've modified the feature makes sense but it may looks overly complex to someone who doesn't know how it used to work.
> 
> If you take a step back, is it how you would build it from the ground up or can you think of a simpler solution?

It makes sense we try to find other ways to simplify here because the font size setting feature itself is simpler now. We used to give to users the ability to change the font size across various parts/components by directly pressing keys (4, 5, 6). Then we learned that was confusing for users. So now we are simplifying by narrowing down the font size setting to one place only and that should be reflected in the code. 

One place to update font size for all subscribed/hooked components, the common denominator being that all components are related to the article reading experience in specific. And we are threading them all with the `.adjustable-font-size` class. Perhaps another solution could be to take advantage of the cascading properties of CSS somehow, set one parent rule and have all other components inherit. But that feels a bit awkward given that all components (`Article`, `ArticlePreview`, `ConfirmDialog`, `QuickFacts`, `ReferencePreview`) are modular as they should be and each has their corresponding stylesheet, so feels a bit against codebase structure. Maybe https://github.com/wikimedia/wikipedia-kaios/pull/290/commits/6f9597da50c0af5765b8b564572e9a87dbbee14f is a step in the right direction. I'd argue we gain on flexibility by keeping the `useArticleTextSize` hooks.

Curious to hear more thoughts though, I'll continue thinking about this.

hueitan (changes requested):
looking good shape now

things to update

1) cleanup the 4,5,6,s,d,f in softkey component
2) according to the design, the blue bar should always fix in the same location when changing the text size
3) design question : do we fine having "Article Menu" previous popup title as the background

stephanebisson (changes requested):
articleTextSize.js line 47 is what I really dislike about this code. This PR is not responsible for it but it might be a good opportunity to clean it up.

Reaching out into the DOM and updating an element in place is the jquery style and pretty much the opposite of the (p)react style. One way to bring this back to the react style would be for useArticleTextSize() to return a class that the consuming components can add to their jsx output. The class can look like `font-size-X` with X being the actual size (12, 13, 14...) or a logical number (1, 2, 3...).

Also, I can't see the point of the FONT_SIZE_ATTRIBUTE now...

style/textsize.less
```diff
@@ -1,5 +1,5 @@
 .textsize {
-	font-size: 17px;
+	height: 130px;
```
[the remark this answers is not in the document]
medied: Yeap, this what it looks like with jio2 screen size (from my browser), few different screenshots to showcase slider/font values

image

I think it looks good, thoughts?

style/menu.less
```diff
@@ -1,4 +1,6 @@
 .menu {
+	height: 160px;
```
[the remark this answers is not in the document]
medied: I kind of like the idea of adding an extra `mode` for popups, specially because this is something that maybe other components can use in the future. I just tested this with a 'halfscreen mode' (50vh), you can see that in this little sub-branch: https://github.com/wikimedia/wikipedia-kaios/commit/c925187f7728705bcba4d9d0e007b91c513688e9

But looking at the results, it seems setting the `height: 160px` directly in `.menu` might be simpler:

- Left column shows how it currently behaves with the `height: 160px` for both portrait and landscape (so without a `@media @landscape` media query). Portrait is in-line with [design mock up](https://app.zeplin.io/project/58dc46f4a83d1e477dd83859/screen/5fc68213375d92a6172724ae) and landscape is readable

- Right column shows the [halfscreen mode](https://github.com/wikimedia/wikipedia-kaios/commit/c925187f7728705bcba4d9d0e007b91c513688e9). Portrait looks basically the same but landscape does look smaller with `50vh`. We could add a media query in this case (`75vh`?) but then that's probably making it more complicated than needed.

Thoughts?

image

style/textsize.less
```diff
@@ -8,11 +8,61 @@
 		box-sizing: border-box;
 		padding: 3px 0;
 		text-align: center;
+		font-size: 17px;
```
hueitan, line 11: probably you already notice the header text doesn't look consistent between device and browser, here's the ticket that we are going to fix, but before going into details, a quick fix is updating the padding to `4px 0` and remove `height`

src/components/TextSize.js
```diff
@@ -1,21 +1,38 @@
...
-import { useI18n, useSoftkey } from 'hooks'
+import { useI18n, useSoftkey, useArticleTextSize } from 'hooks'
+import { articleTextSize } from 'utils'
 
 export const TextSize = ({ close }) => {
   const i18n = useI18n()
 
+  const [textsize, setTextSize] = useArticleTextSize()
+  const { onKeyArrowLeft, onKeyArrowRight } = articleTextSize.getSoftkeyEffect(setTextSize)
+  const sliderValue = ['0', '16.6', '33.2', '49', '66.6', '83.2', '94']
```
hueitan, line 10: when looking at this value on the UI, it seems not very equally distribute in the position, and I can see the difficulty here, so let's think about it after merging.

style/textsize.less
medied: Hmm yea I believe Sudhanshu and I talked about it at some point. Thanks for sharing ticket link though, hadn't seen it yet. I'll tweak padding and remove height though, sounds good

src/components/TextSize.js
medied: Yea I populated `sliderValue` with a mix of math and eye measuring, indeed something we can fine tune later

PR #296, opened 2020-12-11
T267293: font size feature fine tuning

medied:

### Objective

As part of the font size feature [fine tuning](https://phabricator.wikimedia.org/T267293#6682271) and in lieu of the upcoming release, this is an option for what design wants prioritized for the release: having the `TextSize` component be rendered with no popup on the back. This PR aims to have both the CSK and the backspace key return back to article menu while keeping the article menu selected/highlighted state. ~I don't particularly like to query the `document` like that, but wanted to present this solution either way given the release timeline~ _revisited with further commits (#297)_

### Note

Follow up for #290

### DIFF
--- a/style/popup.less
+++ b/style/popup.less
@@ -24,5 +24,9 @@
 			max-height: unset;
 			top: 0;
 		}
+
+		&.hidden {
+			visibility: hidden;
+		}
 	}
 }
--- a/style/popup.less
+++ b/style/popup.less
@@ -24,9 +24,5 @@
 			max-height: unset;
 			top: 0;
 		}
-
-		&.hidden {
-			visibility: hidden;
-		}
 	}
 }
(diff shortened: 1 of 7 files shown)


hueitan (commented):
very bad idea, but i know this is just a temp fix for the release.

medied:
Actually realized we can achieve the same effect with `containerRef`, so moving away from querying `document`

hueitan (approved):
looks good

style/popup.less
```diff
@@ -24,5 +24,9 @@
 			max-height: unset;
 			top: 0;
 		}
+
+		&.hidden {
```
stephanebisson, line 28: What is this for?

medied: Oh good catch, that's leftover from the previous approach so removing...

PR #297, opened 2020-12-11
Hide others

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T267293

### Problem Statement

The article menu popup is partly shown behind the text size popup.

### Solution

Introduce a way for popups to specify that they want to be the only popup visible.


hueitan (approved):
sounds better

medied (approved):
Much better approach, thank you for materializing this idea!

PR #298, opened 2020-12-14
T268751 Use Fulltext search in Search screen

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T268751

### Problem Statement

During search, prefix search always return empty result

### Solution

use fulltext search instead

### Note

well, I can't tell wether the result is better, and <s>can't find the correct prop to return thumbnail</s> ios related code https://github.com/wikimedia/wikipedia-ios/blob/1bc0cea71e2556b60e8b30701e162318f737ffbb/Wikipedia/Code/WMFSearchFetcher.m#L77

### DIFF
--- a/src/api/search.js
+++ b/src/api/search.js
@@ -27,8 +27,9 @@ export const search = (lang, term) => {
       return []
     }
 
-    return Object.values(data.query.search).map(item => {
-      const page = data.query.pages.find(page => page.pageid === item.pageid)
+    const { search, pages } = data.query
+    return Object.values(search).map(item => {
+      const page = pages && pages.find(page => page.pageid === item.pageid)
       return {
         title: item.title,
         description: item.snippet,
--- a/src/api/search.js
+++ b/src/api/search.js
@@ -22,10 +22,24 @@ export const search = (lang, term) => {
   // https://en.wikipedia.org/w/api.php?action=query&list=search&srsearch=no%20result%20page&format=json
   const params = {
     action: 'query',
+    prop: 'pageimages',
+    generator: 'search',
     srprop: 'snippet',
+    srsearch: term,
     srlimit: 15,
     list: 'search',
-    srsearch: term.replace(/:/g, ' '),
+
+    gsrsearch: term,
+    gsrnamespace: 0,
+    // gsrwhat: 'text',
+    // gsrinfo: '',
+    // gsrprop: 'redirecttitle',
+    // gsroffset: 0,
+    gsrlimit: 15,
+    piprop: 'thumbnail',
+    pithumbsize: 64,
+    pilimit: 15,
+
     format: 'json'
   }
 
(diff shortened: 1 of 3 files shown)


hueitan:
one example in english

| Before | After |
| --- | --- |
| image | image |

src/api/search.js
```diff
@@ -1,31 +1,39 @@
...
+    format: 'json'
   }
+
   const url = buildMwApiUrl(lang, params)
   return cachedFetch(url, data => {
-    if (!data.query || !data.query.pages) {
+    if (!data.query || !data.query.search) {
       return []
     }
-    data.query.pages.sort((a, b) => a.index - b.index)
```
hueitan, line 22: @stephanebisson do you still remember why sorting by index here?

src/api/search.js
```diff
@@ -1,31 +1,39 @@
...
+
+    return Object.values(data.query.search).map((p) => {
+      const page = data.query.pages.find(page => page.pageid === p.pageid)
       return {
         title: p.title,
-        titleHtml: p.title.replace(term, `<span class="searchmatch">${term}</span>`),
-        description: p.description,
-        imageUrl: p.thumbnail && p.thumbnail.source
+        titleHtml: p.title,
+        description: p.snippet.replace(/<span class="searchmatch">(\w+)<\/span>/g, '$1'),
```
hueitan, line 35: tbc

searchmatch style still to be confirmed.

src/api/search.js
stephanebisson: It looks like the results have an order based on relevance but they were not ordered inside the json response.

src/api/search.js
```diff
@@ -1,31 +1,39 @@
...
   const url = buildMwApiUrl(lang, params)
   return cachedFetch(url, data => {
-    if (!data.query || !data.query.pages) {
+    if (!data.query || !data.query.search) {
       return []
     }
-    data.query.pages.sort((a, b) => a.index - b.index)
-    return Object.values(data.query.pages).map((p) => {
+
+    return Object.values(data.query.search).map((p) => {
```
stephanebisson, line 30: Not your fault: the parenthesis around the `p` are not needed.

stephanebisson: Maybe that `p` can be renamed since we're looping on search results.

stephanebisson (commented):
Strangely, with this code, when I type `backspace` in the search box, the close confirmation popup appears...

src/api/search.js
hueitan: Thanks, fulltext search has it (relevance) by default, so safe to remove.

src/api/search.js
```diff
@@ -1,31 +1,39 @@
...
+      const page = data.query.pages.find(page => page.pageid === p.pageid)
       return {
         title: p.title,
-        titleHtml: p.title.replace(term, `<span class="searchmatch">${term}</span>`),
```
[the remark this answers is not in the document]
hueitan: we change the search to fulltext search, so this `searchmatch` is previously used for the prefix search to highlight the query string.

src/api/search.js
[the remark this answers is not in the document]
hueitan: so we change to fulltext search, it returns `snippet` instead of the article `description`, to be confirmed that whether we need to highlight the searchterm in snippet.

src/api/search.js
```diff
@@ -1,31 +1,39 @@
 import { cachedFetch, buildMwApiUrl } from 'utils'
 
 export const search = (lang, term) => {
+  // fulltext search
   const params = {
     action: 'query',
-    prop: ['description', 'pageimages', 'pageprops'].join('|'),
+    prop: 'pageimages',
```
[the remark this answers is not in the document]
hueitan: `description` and `pageprops` are for the page item, since changing to `fulltext`, it consider the search response, all we need to get from the prop is only `pageimages`

confirm that we need both `generator` (gs* key used) and `list` (sr* key used)

medied:
@hueitan so this PR is ready for review and not a draft anymore?

stephanebisson (approved):

medied (approved):
Well done!

PR #299, opened 2020-12-15
Update cypress and enable skipped tests

jpita:
Update cypress and enable skipped tests.

NOTE: I changed the docker image running on circleCI to update the browsers where the tests run on, if that's not OK please reject the PR.


hueitan (approved):
looks good to me.

PR #300, opened 2020-12-16
T267293: font size setting refactor

medied:

### Objective

When working on #290, we [spotted a good opportunity for refactoring](https://github.com/wikimedia/wikipedia-kaios/pull/290#pullrequestreview-547779844) which we are carrying out in this PR:

> Reaching out into the DOM and updating an element in place is the jquery style and pretty much the opposite of the (p)react style. One way to bring this back to the react style would be for useArticleTextSize() to return a class that the consuming components can add to their jsx output.


src/hooks/useArticleTextSize.js
```diff
@@ -1,12 +1,18 @@
-import { useState, useLayoutEffect } from 'preact/hooks'
+import { useState, useEffect } from 'preact/hooks'
 import { articleTextSize } from 'utils'
 
-export const useArticleTextSize = (dependencies = []) => {
+let setArticleTextSize
```
stephanebisson, line 4: State like that should be avoided as much as possible in hooks. You'll need a good reason and a good comment to pass code review.

style/fontsizes.less
```diff
@@ -0,0 +1,31 @@
+&.font-size-0 {
```
stephanebisson, line 1: Why the `&` in all those rules?

src/components/TextSize.js
```diff
@@ -20,13 +20,13 @@ export const TextSize = ({ close }) => {
   return <div class='textsize'>
     <div class='header'>{i18n('header-textsize')}</div>
     <div class='content'>
-      <div class='preview adjustable-font-size'>
+      <div class={`preview font-size-${textSize}`}>
```
stephanebisson, line 23: Would it make sense to build the class name inside the hook since this has to be done everywhere?

stephanebisson (commented):
Just sending random thoughts your way. Feel free to ignore while this is a draft.

src/hooks/useArticleTextSize.js
medied: So the reason here is the instantaneous update of the article font size (in the background) as the user clicks left or right in the `TextSize` component with the slider. I tried a couple of different ideas but the behavior always had a delay. That is, the background text wouldn't update instantly and instead update in the next render once the user closed `TextSize` (or even worse after scrolling a bit in the article itself). High level ideas I tried:

* an extra `useEffect` in `ArticleSection` to trigger an `articleTextSize.adjust(0)`, with and without dependencies
* a weird idea that didn't get anywhere was setting the `useArticleTextSize` hook in `ArticleInner`, one level up so that it could be passed down to both `TextSize` and `ArticleSection` and hence use the same state. But alas that didn't work. 
* and then tried different combinations of `useEffect` hooks in `useArticleTextSize`, with and without dependencies, but the only way I was able to get the instant article text update was with that `setArticleTextSize`

Does this spark any better ideas we can try? Idea is to somehow link `ArticleSection` and `TextSize` to achieve the instant update in the background.

src/components/TextSize.js
medied: > Just sending random thoughts your way. Feel free to ignore while this is a draft.

This early feedback is actually useful! Thanks

medied:
> Just sending random thoughts your way. Feel free to ignore while this is a draft.

This early feedback is actually useful! Thanks

src/utils/articleTextSize.js
```diff
@@ -1,8 +1,6 @@
 const KEY = 'article-textsize'
 const DEFAULT_SIZE = 2
 const [MIN_SIZE, MAX_SIZE] = [0, 6]
-const SELECTORS = '.adjustable-font-size'
-const FONT_SIZE_ATTRIBUTE = 'data-font-size'
```
hueitan, line 5: here're some context for you, this attribute is used for **telling the element the initial font size** because every element has different font size in place.

so not just a single `font-size-n` to deal with all elements in article.

src/components/TextSize.js
medied: Yea I think this is a good idea. There's at least one place where the logical integer is needed (in `TextSize` [here](https://github.com/wikimedia/wikipedia-kaios/pull/300/files?file-filters%5B%5D=.js#diff-f841ba9b331ef3886506345a5e90467e92c54d03db86545b229d713a6d70ad67R31-R32)), but I've just adjusted to parse for the integer within that component

src/utils/articleTextSize.js
medied: Right, so I'm proposing we do something slightly different now and take that logic out of attributes and into fontsize.less. With CSS we can adjust the font size as needed for different elements with a combination of `font-size-n` and the corresponding other class:`title`, `desc`, and so on. The `@standard` list in this case is the regular font size used in most cases when `font-size-n` stands alone. The mapping is:


class name | standard (px) | title (px) | preview-title (px) | desc (px)
-- | -- | -- | -- | --
font-size-1 | 14 | 22 | 15 | 12
font-size-2 | 15 | 23 | 16 | 13
font-size-3 | 16 | 24 | 17 | 14
font-size-4 | 17 | 25 | 18 | 15
font-size-5 | 18 | 26 | 19 | 16
font-size-6 | 19 | 27 | 20 | 17
font-size-7 | 20 | 28 | 21 | 18

What do you think?

cypress/support/commands.js
```diff
@@ -11,8 +11,10 @@ import * as enJson from '../../i18n/en.json'
   'backspace',
   'enter'
 ].forEach((key) => {
-  Cypress.Commands.add(key, () => {
-    cy.get('body').type(`{${key}}`)
+  Cypress.Commands.add(key, (repeat = 1) => {
+    for (let i = 0; i < repeat; i++) {
+      cy.get('body').type(`{${key}}`)
+    }
```
medied, line 17: Inspired from https://github.com/wikimedia/wikipedia-kaios/pull/290#discussion_r538947532

src/utils/articleTextSize.js
```diff
@@ -1,8 +1,6 @@
 const KEY = 'article-textsize'
-const DEFAULT_SIZE = 2
-const [MIN_SIZE, MAX_SIZE] = [0, 6]
-const SELECTORS = '.adjustable-font-size'
-const FONT_SIZE_ATTRIBUTE = 'data-font-size'
+const DEFAULT_SIZE = 3
+const [MIN_SIZE, MAX_SIZE] = [1, 7]
```
medied, line 3: Shifted from 0-6 to 1-7 because lists in LESS start counting from 1 (not sure why)

http://lesscss.org/functions/#list-functions-extract

style/fontsizes.less
```diff
@@ -0,0 +1,26 @@
+@standard: 14px, 15px, 16px, 17px, 18px, 19px, 20px;
+@title: 22px, 23px, 24px, 25px, 26px, 27px, 28px;
+@preview-title: 15px, 16px, 17px, 18px, 19px, 20px, 21px;
+@desc: 12px, 13px, 14px, 15px, 16px, 17px, 18px;
```
stephanebisson, line 4: Those are all sequential so you could also have 4 different bases and go `base + value`

style/fontsizes.less
```diff
@@ -0,0 +1,26 @@
...
+
+each(range(7), {
+	.font-size-@{value} {
+		font-size: extract(@standard, @value);
+
+		&.title {
```
stephanebisson, line 10: The font-size is global so you could apply it once, higher up, instead of adding it to every component.

As it stands, the knowledge of which components have adjustable font-size is duplicated to both the components themselves and here. Only one of those is actually needed.

src/utils/articleTextSize.js
```diff
@@ -16,50 +14,29 @@ const set = size => {
...
-  })
-
-  adjust(0)
+  return newSize
 }
 
 const getSoftkeyEffect = (onAdjust = () => {}) => {
-  const onKeyArrowLeft = () => { adjust(-1); onAdjust(get()) }
-  const onKeyArrowRight = () => { adjust(1); onAdjust(get()) }
+  const onKeyArrowLeft = () => { adjust(-1); onAdjust(getFontSizeClassName()) }
```
stephanebisson, line 28: For consistency, Softkey functions should be defined in the component that needs them. They used to be shared, that's why they are here, but that's not the case anymore.

src/hooks/useArticleTextSize.js
```diff
@@ -1,12 +1,18 @@
...
-  }, dependencies)
+export const useArticleTextSize = (origin) => {
+  const [fontSizeClass, setFontSizeClass] = useState(articleTextSize.getFontSizeClassName())
 
-  return [textSize, setTextSize]
+  if (origin === 'Article') {
+    setArticleFontClass = setFontSizeClass
+  }
+
+  useEffect(() => {
```
stephanebisson, line 13: `useEffect` without dependencies will run every time any component using this hook renders. It seems strange that the font size class is set on every render. Shouldn't it be set only when it changes in the UI with the slider?

src/hooks/useArticleTextSize.js
```diff
@@ -1,12 +1,18 @@
-import { useState, useLayoutEffect } from 'preact/hooks'
+import { useState, useEffect } from 'preact/hooks'
 import { articleTextSize } from 'utils'
 
-export const useArticleTextSize = (dependencies = []) => {
-  const [textSize, setTextSize] = useState(articleTextSize.get())
+let setArticleFontClass
```
stephanebisson, line 4: Thanks for the details explanation about why this is there. I understand where you are coming from but sadly that won't do ;)

This is a great opportunity to review why and how the preact components render and to put in place the proper components to make the article text behind the text size popup update in real time.

src/components/Article.js
```diff
@@ -72,7 +72,7 @@ const ArticleSection = ({
     }
   }
 
-  useArticleLinksNavigation('Article', lang, contentRef, linkHandlers, [page, textSize], { galleryItems, articleTitle, namespace, id })
+  useArticleLinksNavigation('Article', lang, contentRef, linkHandlers, [page], { galleryItems, articleTitle, namespace, id })
```
stephanebisson, line 75: `textSize` is needed here because it affects how much text is currently visible on the screen. When it changes, this hook needs to recompute which links are currently visible and can be selected.

stephanebisson (changes requested):
Good job, good improvements but some more to do

hueitan:
just thinking out loud here

as the current solution, it becomes difficult to maintain, let's say if i want to change the article title styling, when looking at `article.style` i don't see the full styles there, but then need to check `fontsize.css` to get the actual size

Would it be possible that we have extra component `<Text />` (or `<AdjustableTextSize/>`) so it observe the context (global textsize state), then we don't have to pass the textsize state into each component.

src/components/Article.js
medied: Yeap, currently working on this issue: when the `useArticleLinksNavigation` hook recomputes it is also triggering a `useSoftkey` update that conflicts with the `TextSize` softkeys as they both use `ArrowLeft`/`ArrowRight` keys

medied: I'm trying out a few ideas around the concept of adding some sort of flag or signal for when the font setting is "complete" (aka: when user presses "OKAY" CSK), but still nothing worth committing yet. This is flag we could use as the dependency for `useArticleLinksNavigation` and hence address the softkey issue. This could also help with the request of separating the functionalities of the backspace key and the CSK.

src/hooks/useArticleTextSize.js
medied: Got it. I'm now switching to using the global context (https://github.com/wikimedia/wikipedia-kaios/pull/300/commits/459b49e94437875efd8a5b6d800414b56d8b5ca5)

src/components/Article.js
```diff
@@ -46,7 +46,7 @@ const ArticleSection = ({
   const i18n = useI18n()
   const [showReferencePreview] = usePopup(ReferencePreview)
   const [showTable] = usePopup(Table, { mode: 'fullscreen' })
-  const [textSize] = useArticleTextSize()
+  const [fontSizeClass] = useArticleTextSize('Article')
```
hueitan, line 49: no longer need to give the origin name here, same to other components.

src/components/TextSize.js
```diff
@@ -5,9 +5,12 @@ import { articleTextSize } from 'utils'
 export const TextSize = ({ close }) => {
   const i18n = useI18n()
 
-  const [textsize, setTextSize] = useArticleTextSize()
-  const { onKeyArrowLeft, onKeyArrowRight } = articleTextSize.getSoftkeyEffect(setTextSize)
+  const [fontSizeClass, setFontSizeClass] = useArticleTextSize('TextSize')
+  const { onKeyArrowLeft, onKeyArrowRight } = articleTextSize.getSoftkeyEffect(setFontSizeClass)
   const sliderValue = ['0', '16.6', '33.2', '49', '66.6', '83.2', '94']
+  const parseTextSizeInteger = (fontSizeClass) => {
```
hueitan, line 11: hmmm i don't understand this line, can you explain a bit?

medied: The purpose of this method is simply to grab number "x" (as integer) of the "font-size-x" class name, which I use to access the appropriate element of the `sliderValue`

src/components/Article.js
medied: Building from the idea above, how about https://github.com/wikimedia/wikipedia-kaios/pull/300/commits/11c9d492c044d9d2cefdb2133442cda3456e32d2? Seems to work well, although I'm trying to understand if there's a better/cleaner way.

medied: Came up with an idea that works better: decoupling the link navigation recomputing from the actual softkeys setting (https://github.com/wikimedia/wikipedia-kaios/pull/300/commits/4f7a3dcc59b4ca0bba68afa267fc67d4189177b0).

style/fontsizes.less
medied: Hmm... once you see the redundancy, you can't unsee it. How about https://github.com/wikimedia/wikipedia-kaios/pull/300/commits/1c945b4d5b7bec571883c0528cf6cd7275b264cd? Using `document.body.className` instead of `classList.add` takes over the `body` class name, I do this to replace font size class as it changes (instead of adding more, `font-size-1 font-size-2` etc). If we ever wanted to another class to the `body` for some reason, we would have to take a look at this part

src/components/App.js
```diff
@@ -18,17 +20,23 @@ export const App = ({ i18n, dir }) => {
   }, [dirState])
   // end of useDirection
 
+  useEffect(() => {
+    document.body.className = fontSizeClass
+  }, [fontSizeClass])
+
```
hueitan, line 26:
```suggestion
  // useFontSize
  const [fontSizeClass, setFontSizeClass] = useState(articleTextSize.getFontSizeClassName())
  useEffect(() => {
    document.body.className = fontSizeClass
  }, [fontSizeClass])
  // useFontSize
```

can we do this scope same as `useDirection`, so we know this scope belongs to text size related code

src/utils/articleTextSize.js
```diff
@@ -16,50 +14,29 @@ const set = size => {
...
-      const fontSize = parseFloat(window.getComputedStyle(element).getPropertyValue('font-size'))
-      element.setAttribute(FONT_SIZE_ATTRIBUTE, fontSize)
-    }
-  })
-
-  adjust(0)
+  return newSize
 }
 
 const getSoftkeyEffect = (onAdjust = () => {}) => {
```
hueitan, line 27: this longer need here, right? as this is done in `TextSize.js` component as @stephanebisson in the comment

src/hooks/useArticleTextSize.js
```diff
@@ -1,12 +1,13 @@
...
-  useLayoutEffect(() => {
-    articleTextSize.init()
-  }, dependencies)
+  useEffect(() => {
+    setFontSizeClass(articleTextSize.getFontSizeClassName())
+  }, [fontSizeClass])
 
-  return [textSize, setTextSize]
+  return [fontSizeClass, setFontSizeClass]
 }
```
hueitan, line 13: this hook can be fully removed as it does return the same thing from `useContext`, and the font size class name is being modified in `app.js`

style/fontsizes.less
```diff
@@ -0,0 +1,29 @@
...
+		.preview-title {
+			font-size: unit(@preview-title + @value, px);
+		}
+
+		.desc,
+		.reference-preview {
+			font-size: unit(@desc + @value, px);
+		}
+	}
+});
```
hueitan, line 29: another way of not generating these css value is using `em` for the relative font size from it's container (parent element), `<AdjustableTextSize/>` component has advantage here

```html
<div class="container">
    <AdjustableTextSize> bla bla bla </AdjustableTextSize>
</div>
```

assume container has `font-size: 10px`
default adjustableTextSize element has `font-size: 1em`

when textsize `adjust(1)` called, adjustableTextSize element change it's font-size to 1.1em; same `adjust(-1)` then `0.9em`

hueitan: so the solution is like 

**CSS** `to handle all font size value` versus **JS** `useEffect( x, dependencies) listener to change element font size em`

src/components/App.js
medied: Sure thing

src/utils/articleTextSize.js
medied: You're right, good catch

src/hooks/useArticleTextSize.js
medied: I see... so just using the `FontContext` directly, I think you're right, let me try that

medied: So check out https://github.com/wikimedia/wikipedia-kaios/pull/300/commits/592781eabb5107738f2b1fbb975443313d505c27. I think the only difference I'm noticing is that useArticleLinksNavigation is _slightly_ off now. Is it just me or do you notice that as well? Looking at `fontSizeClass` in Article.js and App.js, they seem to be updating as expected. But maybe there's a discrepancy with the `fontSizeClass` dependency in Article.js.

Haven't removed useArticleTextSize completely yet for testing purposes

src/components/Article.js
```diff
@@ -1,22 +1,23 @@
...
   usePopup, useTracking
 } from 'hooks'
 import { articleHistory, confirmDialog, goto, viewport } from 'utils'
+import { FontContext } from 'contexts'
 
 const ArticleBody = memo(({ content }) => {
   return (
     <div
-      class='article-content adjustable-font-size'
+      class={'article-content'}
```
stephanebisson, line 20: You don't need an expression between `{}` here because the class is a string literal. 

This applies in many places in this PR.

src/components/App.js
```diff
@@ -18,17 +19,26 @@ export const App = ({ i18n, dir }) => {
   }, [dirState])
   // end of useDirection
 
+  // useFontSize
+  const [fontSizeClass, setFontSizeClass] = useState(articleTextSize.getFontSizeClassName())
+  useEffect(() => {
+    document.body.className = fontSizeClass
```
stephanebisson, line 25: To me, this is the only place where the concept of a "CSS class" is relevant. You could build it here and leave the rest of the code care about the logical values for font sizes (1...7) but not the class itself.

src/components/QuickFacts.js
```diff
@@ -47,13 +46,13 @@ export const QuickFacts = ({ article, goToArticleSubpage, dir, close, closeAll }
     article.contentLang,
     containerRef,
     linkHandlers,
-    [scrollPosition, textSize],
+    [scrollPosition],
```
stephanebisson, line 49: Not sure why this has been removed, changing the font size does change which links are currently visible so it needs to recompute.

src/components/TextSize.js
```diff
@@ -1,32 +1,62 @@
...
+    const currentSizeInteger = parseTextSizeInteger(articleTextSize.getFontSizeClassName())
+
+    articleTextSize.adjust(originalSizeInteger - currentSizeInteger)
+    setFontSizeClass(articleTextSize.getFontSizeClassName())
+    close()
+  }
+
   const sliderValue = ['0', '16.6', '33.2', '49', '66.6', '83.2', '94']
+  const parseTextSizeInteger = (fontSizeClass) => {
+    return parseInt(fontSizeClass.slice(-1))
```
stephanebisson, line 39: Following an earlier comment, the class should be built as needed based on the font size, the font size should not have to be extracted from the class.

stephanebisson (commented):
I haven't tested. I will do that later.

src/components/Article.js
medied: Good catch, this is left over from when we were placing `fontSizeClass` in each element directly

src/components/QuickFacts.js
medied: This was needed before because the user was able to change the font size from within `Quickfacts` by directly pressing 4, 5, 6 but now that's not the case anymore

hueitan (changes requested):
As mention in the chat, there are 2 bugs found

**(1)** useArticleLinksNavigation observation doesn't work correctly
1. go to article `Cat`
1. change the text size to the largest first
1. go to section `Etymology_and_naming`
1. then change the text size to smallest
1. arrrow right but unable to select the bottom two articles

**(2)** softkey fail when opening gallery from quickfact
1. go to article `Cat`
1. go to quickfact
1. press enter to open gallery
1. the softkey is error there

src/components/TextSize.js
```diff
@@ -1,32 +1,56 @@
...
+
   useSoftkey('TextSize', {
     center: i18n('softkey-ok'),
     onKeyCenter: close,
-    onKeyBackspace: close,
-    onKeyArrowLeft,
-    onKeyArrowRight
+    onKeyBackspace,
+    onKeyFixedArrowLeft,
+    onKeyFixedArrowRight
```
hueitan, line 41: These we should use `onKeyArrowLeft` and `onKeyArrowRight`

FixedArrow is not this purpose

medied: What is the difference between `onKeyArrowRight/Left` and `onKeyFixedArrowRight/Left`?

medied:
>(1) useArticleLinksNavigation observation doesn't work correctly

What I'm seeing is that `findVisibleLinks` is triggering too early, it's giving us the amount of visible links right before the font size changes

>(2) softkey fail when opening gallery from quickfact

https://github.com/wikimedia/wikipedia-kaios/pull/300/commits/eee6bdc12e0baf7722abeaea7b1b9fe8bb9e77ee

src/hooks/useArticleLinksNavigation.js
```diff
@@ -57,6 +60,15 @@ export const useArticleLinksNavigation = (
...
   }, dependencies)
 
+  useEffect(() => {
+    const hasGallery = popupState.find(p => p.id === 'Gallery')
+    if (!hasGallery) {
+      if (origin !== 'Article' || popupState.length === 0) {
+        setSoftkeyTrigger(!softkeyTrigger)
+      }
+    }
+  }, [popupState, currentLink])
```
hueitan, line 70: This is the addition line that I miss before, could you explain what happen here?

src/components/TextSize.js
hueitan: `onKeyArrowRight/Left` will be switched if `direction=rtl`

`onKeyFixedArrowRight/Left` doesn't no matter the `direction` value, one example is `Table.js`

medied: Gotcha that makes sense, thanks. Will update handlers to `onKeyArrowRight/Left` soon

src/hooks/useArticleLinksNavigation.js
medied: Originally, the first `useEffect` in `useArticleLinksNavigation` would watch the `dependencies` passed in to recompute visible links and then consequently trigger the `useSoftkey` hook by updating `links` and `currentLink`. This was fine when changing fontsize direcly with numberic keys (4, 5, 6) but now results in a softkey bug.

After trying a couple of things, the idea I'm proposing is [to decouple](https://github.com/wikimedia/wikipedia-kaios/pull/300#discussion_r551890900) the recomputing of visible links from the actual softkeys setting, and in order to do that I'm introducing this second `useEffect` that decides when to toggle the `softkeyTrigger` by observing the `popupState`

hueitan (changes requested):
Looks good now, two more things to go before qa test

1. merge latest master
2. apply it to suggested articles (article footer)

hueitan (approved):
Looks good! @jpita when testing please double check against the changes part (mostly the entire article ;-)

src/components/TextSize.js
```diff
@@ -1,36 +1,64 @@
...
-import { useContext } from 'preact/hooks'
-import { useI18n, useSoftkey, useArticleTextSize } from 'hooks'
+import { useEffect, useContext } from 'preact/hooks'
+import { useI18n, useSoftkey } from 'hooks'
 import { articleTextSize } from 'utils'
-import { DirectionContext } from 'contexts'
+import { FontContext, DirectionContext } from 'contexts'
 
-export const TextSize = ({ close }) => {
+let originalTextSize
```
stephanebisson, line 7: Sorry, please don't do that.

If you need to store state, there's `useState`.

src/utils/articleTextSize.js
```diff
@@ -1,8 +1,12 @@
...
+ */
+
 const KEY = 'article-textsize'
-const DEFAULT_SIZE = 2
-const [MIN_SIZE, MAX_SIZE] = [0, 6]
-const SELECTORS = '.adjustable-font-size'
-const FONT_SIZE_ATTRIBUTE = 'data-font-size'
+const DEFAULT_SIZE = 3
+const [MIN_SIZE, MAX_SIZE] = [1, 7]
+let hasAdjusted = false
```
stephanebisson, line 9: Implicit state outside of a util definition is not allowed.

style/fontsizes.less
```diff
@@ -0,0 +1,40 @@
...
+		.title {
+			font-size: unit(@title + @value, px);
+		}
+
+		.preview-title {
+			font-size: unit(@preview-title + @value, px);
+		}
+
+		.footer-title {
+			font-size: unit((@desc + @value) + 0.8, px);
```
stephanebisson, line 30: Why not have a footer base value like for all the other ones?

src/components/Article.js
```diff
@@ -72,7 +71,7 @@ const ArticleSection = ({
     }
   }
 
-  useArticleLinksNavigation('Article', lang, contentRef, linkHandlers, [page, textSize], { galleryItems, articleTitle, namespace, id })
+  useArticleLinksNavigation('Article', lang, contentRef, linkHandlers, [page, textSizeAdjustment], { galleryItems, articleTitle, namespace, id })
```
stephanebisson, line 74: Why don't you depend on the text size itself here instead of the "adjusted" variable. Typically, it's `useEffect`'s job to figure out if dependencies have changed between cycles.

stephanebisson (changes requested):
Making good progress but those floating `let` variables are red flags.

src/components/Article.js
medied: It’s because as `textSize` state changes instantly when the users plays with the slider, the `useEffect` in `useArticleLinksNavigation` triggers every time and consequently sets the softkeys at the wrong moment. Hence the `hasAdjusted` state to link `TextSize` and `Article`. We could also just have the text size slider update the preview text in `TextSize` only and not the article text on the background, that way we could use `textSize` as the dependency directly

src/components/TextSize.js
```diff
@@ -1,36 +1,59 @@
...
-import { DirectionContext } from 'contexts'
+import { FontContext, DirectionContext } from 'contexts'
 
-export const TextSize = ({ close }) => {
+export const TextSize = ({ close, closeAll }) => {
   const i18n = useI18n()
   const { dirState } = useContext(DirectionContext)
-  const [textSize, setTextSize] = useArticleTextSize()
+  const { textSize, setTextSize } = useContext(FontContext)
+  const originalTextSize = articleTextSize.get()
```
stephanebisson, line 11: This will run every time so there is no guarantee that it will contain the true original value.

To initialize it on the first run and keep it for the lifecycle of the component you can do:

`const [originalTextSize] = useState(articleTextSize.get())`

style/fontsizes.less
```diff
@@ -0,0 +1,41 @@
...
+		.title {
+			font-size: unit(@title + @value, px);
+		}
+
+		.preview-title {
+			font-size: unit(@preview-title + @value, px);
+		}
+
+		.footer-title {
+			font-size: unit((@footer-title + @value) + 0.8, px);
```
stephanebisson, line 31: That `@footer-title` could include the additional 0.8

src/components/Article.js
stephanebisson: Right, I remember this issue now. And looking at the code, a clean solution would involve `useSoftkeys` understanding that it should update the handlers for a component that is not top-most at the moment. Looking at the SoftkeyReducer it seems possible but it adds scope when we should be in wrap up mode here.

Let's talk about this later.

hueitan (approved):
Code looks clean in the local text size change approach 🎉

src/utils/articleTextSize.js
```diff
@@ -1,8 +1,11 @@
+/**
+ * The font size state is set to the HTML body to be made available
+ * for all elements with adjustable text size. See fontsize.less for more info
+ */
+
 const KEY = 'article-textsize'
```
stephanebisson, line 6: The storage key is unchanged but the values it can contain have changed from [0,6] to [1,7].

How would the new code react when it reads the value 0? 

I see a few ways of handling that:
1. On read, validate the value and migrate it (0 becomes 1)
2. Version the key (KEY = 'article-textsize-v2')
3. Keep the stored variable domain as [0,6] and add 1 when you build the class so that everybody's setting is retained.

stephanebisson (changes requested):
Looks good. One last thing...

src/utils/articleTextSize.js
medied: Hmm good point, all of these ideas are fair. Sort of leaning towards the third option... let me try it out

stephanebisson (approved):

PR #302, opened 2021-01-07
show the equal portion for each slider value

hueitan:
Phabricator Link: -

### Problem Statement

The visual UI of the text size component doesn't feel right

### Solution

calculate the portion to get the nearest value

### Note

merge to refactor branch

### DIFF
--- a/src/components/TextSize.js
+++ b/src/components/TextSize.js
@@ -2,22 +2,23 @@ import { h } from 'preact'
 import { useEffect, useContext } from 'preact/hooks'
 import { useI18n, useSoftkey } from 'hooks'
 import { articleTextSize } from 'utils'
-import { FontContext } from 'contexts'
+import { FontContext, DirectionContext } from 'contexts'
 
 let originalTextSize
 
 export const TextSize = ({ close }) => {
   const i18n = useI18n()
+  const { dirState } = useContext(DirectionContext)
   const { textSize, setTextSize } = useContext(FontContext)
   const sliderPortion = 100 / 6
   const sliderValue = Array.from({ length: 7 }, (v, i) => i * sliderPortion)
 
-  const onKeyFixedArrowLeft = () => {
+  const onKeyArrowLeft = () => {
     articleTextSize.adjust(-1)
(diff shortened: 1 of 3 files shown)


src/components/TextSize.js
```diff
@@ -9,7 +9,8 @@ let originalTextSize
 export const TextSize = ({ close }) => {
   const i18n = useI18n()
   const { textSize, setTextSize } = useContext(FontContext)
-  const sliderValue = ['0', '16.6', '33.2', '49', '66.6', '83.2', '94']
+  const sliderPortion = 100 / 6
+  const sliderValue = Array.from({ length: 7 }, (v, i) => i * sliderPortion)
```
hueitan, line 13: Question here: what about this constant 7 , we get it from the articleTextSize util?

so the constant can be reused for the value of portion and array length.

medied: You mean like having a variable that we can reuse for both creating the `sliderValue` and in `articleTextSize` util? `MAX_SIZE`?

medied: Neat idea by the way, I think it looks a bit smoother

style/textsize.less
```diff
@@ -50,6 +50,7 @@
 				top: -6px;
 				border: solid 3px #fff;
 				background-color: #36c;
+				transform: translateX( -50% );
```
medied, line 53: 👌

hueitan:
@medied this is done + direction handling.

src/components/TextSize.js
```diff
@@ -37,20 +40,20 @@ export const TextSize = ({ close }) => {
     center: i18n('softkey-ok'),
     onKeyCenter: close,
     onKeyBackspace,
-    onKeyFixedArrowLeft,
-    onKeyFixedArrowRight
+    onKeyArrowLeft,
+    onKeyArrowRight
```
medied, line 44: Oh thanks for updating these!

medied:
> @medied this is done + direction handling.

Awesome, thanks! Will test and merge now then

PR #306, opened 2021-01-13
T271359: add undefined guard to isImageInGallery

medied:
https://phabricator.wikimedia.org/T271359

### Objective

When an [article has no images](http://localhost:8080/#/article/fr/Veto), the gallery array is provided as an empty array (as opposed to `undefined`). The one case I can spot where `galleryItems` is provided to `useArticleLinksNavigation` as `undefined` is when invoked via `ReferencePreview`. I haven't found yet a specific example of a reference that replicates the issue, it would have to be a reference that contains a `figure` or `figure-inline` tag (is that even possible?). Also, the links navigation does not break for either `ReferencePreview` or `Article` as far as I can tell

### DIFF
--- a/src/hooks/useArticleLinksNavigation.js
+++ b/src/hooks/useArticleLinksNavigation.js
@@ -173,7 +173,7 @@ const findVisibleLinks = (container, galleryItems) => {
 
 const isImageInGallery = (galleryItems, link) => {
   const aElement = link.querySelector('a')
-  if (!aElement) {
+  if (!aElement || !galleryItems) {
     return false
   }
 


stephanebisson (approved):

src/hooks/useArticleLinksNavigation.js
```diff
@@ -173,7 +173,7 @@ const findVisibleLinks = (container, galleryItems) => {
 
 const isImageInGallery = (galleryItems, link) => {
   const aElement = link.querySelector('a')
-  if (!aElement) {
+  if (!aElement || !galleryItems) {
```
hueitan, line 176: never try but it is possible galleryItems is `[]` ?

medied: `galleryItems` is `[]` in article with no images like http://localhost:8080/#/article/fr/Veto

jpita: @medied if I do either

``` 
const isImageInGallery = ([], link) => {
  const aElement = link.querySelector('a')
```

or 

```
const isImageInGallery = (galleryItems, link) => {
  const aElement = []
```

the navigation is broken (can't use any buttons or leave the section) when the image in the "Attack" section is loaded in http://0.0.0.0:8080/#/article/en/2010_Bratislava_shooting.

is either of the outcomes possible with ur code?

hueitan: no

the first code is invalid, the second code `aElement` will never be empty array

> querySelector returns either element or null

PR #307, opened 2021-01-18
Add source-map for local debugging

stephanebisson:
Phabricator Link: -

### Problem Statement

Sometimes you need to step through the code.

### Solution

Add source-map to view the source in the browser debugger as written.


hueitan (approved):
Nice one! so we generate the source-map for the build and use to deploy the local device!

medied (approved):
I think it works! Seeing more useful info on the browser when setting a `debugger` breakpoint, thanks for adding this

PR #309, opened 2021-01-19
T270469: onboarding update

medied:
https://phabricator.wikimedia.org/T270469

### Objective

Update`Onboarding` so that it's reduced to one screen and update `Consent` screen to be displayed as with half-screen height with scrolling. Users would still need to agree to our terms and services before start using the app.

image

### Note

- [x] Please don't merge until I sync with Sudhanshu on some styling details.


src/components/Onboarding.js
```diff
@@ -1,43 +1,30 @@
...
+        {i18n('onboarding-0-title')}
       </div>
       <div class='description'>
-        {i18n(`onboarding-${currentIndex}-description`)}
-      </div>
-      <div class='indicator'>
-        <div class={`dot ${currentIndex === 0 ? 'selected' : ''}`} />
-        <div class={`dot ${currentIndex === 1 ? 'selected' : ''}`} />
-        <div class={`dot ${currentIndex === 2 ? 'selected' : ''}`} />
+        {i18n('onboarding-0-description')}
```
medied, line 27: Considered renaming the `onboarding-0` to just `onboarding` but `AboutWikipedia` uses the same keys so opted to leave as is

src/utils/consent.js
```diff
@@ -27,7 +27,8 @@ export const consentMessages = {
     'consent-message-terms': 'I understand and accept the Terms of Service.',
     'consent-softkeys-policy': 'Policy',
     'consent-softkeys-agree': 'Agree',
-    'consent-softkeys-terms': 'Terms'
+    'consent-softkeys-terms': 'Terms',
+    'consent-privacy-terms': 'Privacy and Terms'
```
medied, line 31: Added it here instead of en/qqq.json to leave all consent message grouped together

hueitan (changes requested):
should remove all images/onboarding-* related resource

src/components/Routes.js
```diff
@@ -1,22 +1,19 @@
...
-import { onboarding, isConsentGranted, goto } from 'utils'
+import { Article, Search, Settings, Language, Onboarding } from 'components'
+import { onboarding } from 'utils'
 
 export const Routes = ({ onRouteChange }) => {
   const onChange = ({ url }) => {
     if (!onboarding.isDone()) {
       route('/onboarding')
-    } else if (!isConsentGranted()) {
-      goto.consent()
```
hueitan, line 12: don't forget to remove consent() in goto.js

stephanebisson:
> should remove all images/onboarding-* related resource

Are they not still being used in "About Wikipedia"?

src/utils/consent.js
stephanebisson: This is just for messages that have legal implications and that are translated by legal experts. This is not hooked into TWN so please move this key to en/qqq.

hueitan:
right! I forgot about there is `About page`

stephanebisson (changes requested):
I think we have to make sure the instrumentation code is not running if the consent is not granted yet. We could return early in `useTracking` or even in `sendEvent` in that case.

Also, when I give consent, the focus doesn't go into the search box. That's in the browser, I haven't tested on a device yet.

medied:
Interestingly, updating `useTracking` [like this](https://github.com/wikimedia/wikipedia-kaios/pull/309/commits/186f2c071b43940835aca431306b458e5e068525) makes it so that the `useEffect` in `Search` is triggered again on granting consent, therefore placing the cursor in the search bar. Tested in both browser and device 🍌

src/components/Consent.js
```diff
@@ -23,14 +26,16 @@ export const Consent = () => {
       onKeyLeft: goto.termsOfUse,
       right: i18n('consent-softkeys-policy'),
       onKeyRight: goto.privacyPolicy,
+      onKeyArrowDown: scrollDown,
+      onKeyArrowUp: scrollUp,
       backspace: () => { window.exit() }
```
hueitan, line 31: this should be `onKeyBackspace`, i don't know why we use backspace

and it should also `window.close()`, not `window.exit()` ... 😨

hueitan:
Offline status UI, previouly was fullscreen, need to ask design to think about it 
Screen Shot 2021-01-21 at 19 13 25

src/components/Search.js
```diff
@@ -77,8 +79,12 @@ export const Search = () => {
...
   useEffect(() => {
     articleHistory.clear()
-    setNavigation(0)
-  }, [])
+    if (!consentGranted) {
+      showConsentPopup()
+    } else {
+      setNavigation(0)
+    }
+  }, [consentGranted])
```
hueitan, line 87: `consentGranted` is not a state, just a variable, should be fine just use it in the effect

src/components/Consent.js
medied: Good catch!

src/components/Search.js
medied: When I test on device with the `useEffect` below, the result I'm seeing is that the cursor/focus is not placed on search bar after consent. 

```
useEffect(() => {
    articleHistory.clear()
    if (!isConsentGranted()) {
      showConsentPopup()
    } else {
      setNavigation(0)
    }
  }, [])
```

With `consentGranted` in the dependency array I do see the focus on search bar after grating consent on device

hueitan: also instrumentation code works only when the dependency array has `consentGranted`

🤔 don't understand why the `consentGranted` is not either prop or state but works

src/components/Consent.js
stephanebisson: Wait, backspace on the consent screen should indeed exit the app since you're not allowed to use it unless you explicitly agree to the terms.

medied: Yea I tested on device after https://github.com/wikimedia/wikipedia-kaios/pull/309/commits/a3ab443e803a18c1022dae8869af3e694b8f6f93 and I'm seeing that the app exits as expected, then on reopening the consent screen is displayed again

stephanebisson:
Works well but that half-visible offline screen is a deal breaker.

medied:
Checking in with design regarding the offline view, stay tuned...

src/components/Search.js
medied: A bit puzzling to me as well actually, as I understand it `consentGranted` is acting as state since `isConsentGranted` reads from `localStorage` after `onAgree` handler runs in `Consent`

src/components/Consent.js
```diff
@@ -1,44 +1,46 @@
...
+    onKeyBackspace: () => { window.close() }
+  })
+
+  useEffect(() => {
+    if (!isOnline) {
+      close()
     }
-  }
-  useSoftkey('ConsentMessage', softkeyConfig[isOnline ? 'online' : 'offline'], [isOnline], true)
+  }, [isOnline])
```
medied, line 33: Reasoning and test I did for this here:

* installed app on device
* wifi is on
* open app and get to consent screen as expected
* leave wifi on and app open and walked two floors down the stairway until connection was lost (rather than closing the app and turning wifi off manually)
* as connection was lost, the app closed consent screen and displayed offline panel

hueitan: the above step is valid and make sense, and user can continue with

1. on the offline screen, type query (the app doesn't send the api request)
2. once the app goes online, the app shows consent screen correctly, but in the background it loads search data 

😅

hueitan, line 33: This logic looks weird and doesn't make sense and the controller of the online/offline state is in the upper level (search screen)

I propose `usePopup` to return `close` method, then within the search component we use

```js
useEffect(() => {
    // ... other code
    if (!isOnline) {
      closeConsentPopup()
    }
    // ... other code
  }, [isOnline])
```

medied: Good catch! What about https://github.com/wikimedia/wikipedia-kaios/pull/309/commits/cc18e0ed7bce2a68aa2d53ac2ed4c2844e0bdf77 to disable input in that edge case?

src/components/Search.js
```diff
@@ -64,6 +66,13 @@ export const Search = () => {
     })
   }
 
+  const allowInput = () => {
+    if (isOnline === false && consentGranted === false) {
+      return false
+    }
+    return true
```
hueitan, line 73:
```suggestion
    return isOnline || consentGranted
```

simplify

online : always show the input
offline but has consent : show input
offline but no consent : hide input

src/components/Search.js
```diff
@@ -88,6 +88,8 @@ export const Search = () => {
     articleHistory.clear()
     if (!consentGranted && isOnline) {
       showConsentPopup()
+    } else if (!isOnline) {
+      closeConsentPopup()
```
hueitan, line 92: this case `has consent and offline` should also focus on input

src/components/Search.js
```diff
@@ -88,6 +88,8 @@ export const Search = () => {
...
     } else {
       setNavigation(0)
     }
```
hueitan: actually just

```js
} else {
  closeConsentPopup();
  setNavigation(0);
}
```

should work, right?

medied: I think you're right!

src/components/Search.js
medied: Love it

src/components/Search.js
```diff
@@ -64,26 +68,35 @@ export const Search = () => {
...
-    right: i18n('softkey-settings'),
-    onKeyRight: () => { window.location.hash = '/settings' },
+    right: allowUsage() ? i18n('softkey-settings') : '',
+    onKeyRight: allowUsage() ? () => { window.location.hash = '/settings' } : null,
     center: current.type === 'DIV' ? i18n('centerkey-select') : '',
     onKeyCenter,
     onKeyLeft: isRandomEnabled() ? goToRandomArticle : null,
     onKeyBackspace: !(query && current.type === 'INPUT') && onExitConfirmDialog
-  }, [current.type, query])
+  }, [current.type, query, isOnline])
```
medied, line 82: I think we should merge this PR before #313, so that I can then resolve conflicts in the second PR which would clean this up a bit

hueitan (approved):
Looks good now! 👍

medied:
@jpita when QAing this PR, let's heavy test that the user is never able to use the search bar before granting consent.

PR #312, opened 2021-01-21
Update version in Manifest file

hueitan:
Phabricator Link: -

### Problem Statement

the build script will update manifest version using package.json, but we should also update it within the manifest file, so we get the correct version number during development.

### Solution

update version in manifest


stephanebisson:
Trying to keep it in sync in 2 places is annoying. What would be the benefit?

hueitan:
right, it's annoying, I don't see a better way unless just remove it from the manifest file.

the only benefit I can think of right now is the instrumentation testing, as we always sent the version `1.0.0` which is not the real latest version.

stephanebisson:
We could set it to 0.0.0 in the manifest so we know for sure this is coming from local testing and not any deployed version.

hueitan:
> We could set it to 0.0.0 in the manifest so we know for sure this is coming from local testing and not any deployed version.

make sense for me!

medied (approved):

PR #314, opened 2021-01-29
Various fixes for iframe and Chrome

stephanebisson:
Phabricator Link: -

In the context of building a KaiOS simulator, various fixes were needed to make the app play nice inside an iframe and in the Chrome browser.

* Use `window.history` to access the history specific to the current frame
* Encapsulate `history.back()` in the `goto` utils
* Create card transparency using the rgba filter because `opacity` breaks pagination in Chrome
* Fix XHR error handling when the response object is null (it happens in Firefox when there is CORS issues)
* Use the height/width from the current window or frame and not the entire screen
* Don't mangle variable names during minification
* Turn `viewport` into a function so it gets the right value after a resize (during tests)


hueitan (approved):
I would like to give this PR update 💯 🥇 

@jpita so finally, this pagination works in Chrome with this PR!

jpita:
> I would like to give this PR update 💯 🥇
> 
> @jpita so finally, this pagination works in Chrome with this PR!

yes! so good!
we can run the tests headlessly with eletron now, should I make that change?

stephanebisson:
> we can run the tests headlessly with eletron now, should I make that change?

Unless there is real benefits to using electron, Firefox is still closer to our target environment.

jpita:
> Unless there is real benefits to using electron, Firefox is still closer to our target environment.

not a huge benefit, it just runs 20/30 seconds faster.
not too important IMO

PR #315, opened 2021-02-01
update package-lock.json

hueitan:
update it as no idea why they are so many difference from the latest `npm i`


stephanebisson:
Your setup downgraded the package lock to `"lockfileVersion": 1,`

stephanebisson:
Let's not convert the package lock from version 2 to version 1.

PR #316, opened 2021-02-01
Trim search query to prevent searching for empty string

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T272276

### Problem Statement

Searching for empty string returns no results

### Solution

Trim the search query and only search if it is not empty

### DIFF
--- a/src/components/Search.js
+++ b/src/components/Search.js
@@ -21,7 +21,7 @@ export const Search = () => {
   const [query, setQuery, searchResults] = useSearch(lang)
   const isOnline = useOnlineStatus(online => {
     if (online) {
-      setQuery(inputRef.current.value)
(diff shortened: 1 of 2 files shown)


hueitan:
I think these two scenarios are fine

1. Type `   `, then hit back key, it shows exit confirm dialog, cancel the dialog will remove all `   `
1. Type `   `, then type another digit `a`, will clear previous  `   `

hueitan:
what about this case

1. Type `president obama`
2. user want to research for `president xxx`, then he deletes `obama`
3. Once the cursor reach `o`, hit delete will bring the cursor to `t`, not the empty space there

hueitan (approved):

PR #318, opened 2021-02-15
Assign users into group and add to instrumentation

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T271319

### Problem Statement

As part of the KaiOS engagement experiment, users have to be divided into groups and we have to know which group they are in for analysis.

### Solution

Always include the experiment name and the user's group in the InukaPageView instrumentation calls. The group is stored in localStorage and determined randomly the first time it is accessed.

Also include getters for specific groups.

### Note

There isn't much to observe here. This is foundational to other PRs that will bring the UI of the experiment.


medied (approved):
I can start using this already in the PR I'm working on, will be on the lookout for the merge

hueitan:
Is this expected that the experiment group is set after one instrumentation code sent? 

that means for the user, they won't be able to see the experiment first when landing first screen (given that they will be assigned to experiment group later) but after go to either settings or one article page

stephanebisson:
> Is this expected that the experiment group is set after one instrumentation code sent?

It should never be the case. Are you able to reproduce it?

PR #320, opened 2021-02-15
Loading state when searching

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T271916

### Problem Statement

On a slow connection, it's not always clear that a search is in progress or if the user is expected to click somewhere to trigger it.

### Solution

Provide a loading state on the search page. It is the same as the article loading state but with a different message.

### Note

Also tried to re-organize when what should be shown in the search page since that was getting a little confusing. I hope it helps.


src/components/Search.js
```diff
@@ -93,12 +93,19 @@ export const Search = () => {
     }
   }, [consentGranted, isOnline])
 
+  const hideW = (searchResults || !isOnline || loading)
+  const showSearchBar = allowUsage()
+  const showResultsList = isOnline && searchResults && !loading
+  const showLoading = isOnline && loading
+  const showOfflinePanel = !isOnline
```
medied, line 100: Love it, was thinking of something similar. This will be useful with tips and feed features coming up

medied (approved):
Much needed

hueitan:
Due to the absolute position, the planet is not running in the center screen while loading the article. 
I'm okay with this if design approve.

| This update | Previously |
| --- | --- |
|  |  |

hueitan:
here the latest screenshot for your reference

| Search loading (Portrait) | Search loading (Landscape) | Article Loading (Portrait) | Article loading (Landscape) |
| --- | --- | --- | --- |
| localhost_8080_ | localhost_8080_ (3) | localhost_8080_ (4) | localhost_8080_ (2) |

hueitan:
I don't understand why the CI environment has different search result with the key `catt`, any chance you see why? @jpita

jpita:
> I don't understand why the CI environment has different search result with the key `catt`, any chance you see why? @jpita
this is what I see on my local env
image

this is what the CI sees
image

they look the same to me.
either the search API results changed or we changed the way we call/use the search results.
if u want you can add a step in the test to go to the 2nd result (catt article )

stephanebisson (commented):
LGTM

I can't approve it since I'm one of the author. @medied, you're our only hope.

hueitan (approved):
😅

medied (approved):
I got you!

PR #322, opened 2021-02-16
T271904: Trending articles feed

medied:
https://phabricator.wikimedia.org/T271904

### Objective 

Add UI for trending articles feed

---

TODO

- [x] Revert https://github.com/wikimedia/wikipedia-kaios/pull/322/commits/7a6ac2b35da43023317c60c2537f5a19262f9a49 after testing


style/feed.less
```diff
@@ -0,0 +1,110 @@
...
+	.cue {
+		width: 15px;
+		height: 3px;
+		margin: auto;
+		margin-top: 8px;
+		border-radius: 1.5px;
+		background-color: #ccc;
+	}
+
+	.list {
```
stephanebisson, line 58: <ListView> already has its own styling consistent across the app. Wondering why we repeat some of it here and it's also slightly different, especially the selected item.

stephanebisson (commented):
Should probably merge master into this PR to clean it up.

style/feed.less
medied: You're right I do need to clean up the repeated rules. I'm overriding the slightly different parts in feed.less so it fits what design is requesting

src/hooks/useNavigation.js
```diff
@@ -18,16 +18,32 @@ export const useNavigation = (origin, containerRef, listRef, axis, elementsSelec
   const navigatePrevious = () => {
     const allElements = getAllElements()
     const currentIndex = getTheIndexOfTheSelectedElement()
+    let setIndex = (currentIndex === 0) ? allElements.length - 1 : currentIndex - 1
+
+    if (origin === 'Search') {
```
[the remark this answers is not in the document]
medied: Yea, let's talk more about this later

i18n/en.json
```diff
@@ -81,6 +81,14 @@
...
     "about-app-message": "Made by the Wikimedia Foundation with the help of volunteers like you.",
     "softkey-read-more": "Read More",
+    "feed-header": "Trending",
+    "feed-topic-entertainment": "Entertainment",
+    "feed-topic-sports": "Sports",
+    "feed-topic-people": "People",
+    "feed-topic-science": "Science",
+    "feed-topic-culture": "Culture",
+    "feed-topic-places": "Places",
+    "feed-topic-society": "Society",
```
stephanebisson, line 91: These keys are in master, they shouldn't be in this diff. Are you up to date?

src/components/Feed.js
```diff
@@ -0,0 +1,24 @@
...
+  useEffect(() => {
+    const [request, abort] = getTrendingArticles(lang)
+    request.then(setTrendingArticles)
+    return abort
+  })
+
+  return (
+    <div class={`feed ${isExpanded ? 'expanded' : 'collapsed'}`}>
+      {!isExpanded && <div class='cue' />}
+      {isTrendingArticlesGroup() && <ListView items={trendingArticles} header={i18n('feed-header')} containerRef={containerRef} />}
```
stephanebisson, line 21: I don't know if `isTrendingArticlesGroup()` is needed here. This is already checked in the Search component and if the Feed component is included it means the user must be in the trending group.

medied: You're right, this will be needed here only when the topics feed will be used

stephanebisson:
Bug: When in the control group, hitting the down button brings the search page in "feed expanded mode" but there's no feed.

medied:
**NOTE**: let's not merge before reverting https://github.com/wikimedia/wikipedia-kaios/pull/322/commits/7a6ac2b35da43023317c60c2537f5a19262f9a49

src/hooks/useNavigation.js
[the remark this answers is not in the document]
medied: How about https://github.com/wikimedia/wikipedia-kaios/pull/322/commits/3a8fb0cac3d3523cbe8ffcca8478681377b2b16f? This way we don't add coupling in the wrong direction and we preserve the same scrolling behavior for search results

src/components/Search.js
```diff
@@ -16,7 +17,8 @@ export const Search = () => {
   const inputRef = useRef()
   const listRef = useRef()
   const i18n = useI18n()
-  const [current, setNavigation, getCurrent] = useNavigation('Search', containerRef, listRef, 'y')
+  const [isFeedExpanded, setIsFeedExpanded] = useState(false)
+  const [current, setNavigation, getCurrent, getAllElements, navigateNext, navigatePrevious] = useNavigation('Search', containerRef, listRef, 'y')
```
hueitan, line 21: just looking at the source, there is a way to cleaning some codes here

`navigateNext` is also `setNavigation(getCurrent().index + 1)`
and 
`navigatePrevious` is also `setNavigation(getCurrent().index - 1)`

stephanebisson (commented):
I'm not too happy about the complexity in the search component but it does mirror the complexity of the feature and there is is no clear violation as far as I can see so I'm ok with it.

I see a strange bug: start typing in the search box and hit the down key before results have come back and look for errors in the console.

medied:
> I see a strange bug: start typing in the search box and hit the down key before results have come back and look for errors in the console.

Good catch, we can observe `loading` to avoid that

cypress/support/commands.js
```diff
@@ -49,6 +49,7 @@ Cypress.Commands.add('clickMenuButton', () => {
 Cypress.Commands.add('navigateToHomePage', () => {
   cy.setLocalStorage('has-onboard-before', true)
   cy.setLocalStorage('usage-data-consent', '{}') // Don't display consent screen
+  cy.setLocalStorage('experiment-group', 'control')
```
hueitan, line 52: the localstorage name isn't this one

src/components/Search.js
```diff
@@ -89,11 +131,12 @@ export const Search = () => {
     }
   }, [consentGranted, isOnline])
 
-  const hideW = (searchResults || !isOnline || loading)
+  const hideW = (searchResults || !isOnline || loading || (isFeedExpanded && isTrendingArticlesGroup()))
```
hueitan, line 134:
```suggestion
  const hideW = searchResults || !isOnline || loading || isFeedExpanded
```

I think this can be short like this, there is no need to double check if it is within the group, the `isFeedExpanded` is enough here

hueitan:
when I go to another language from settings, should it avoid here? or config

stephanebisson:
The experiment is only supposed to run for English. There was a task about restricting it to English and Hindi but with the change of direction it has disappeared from the board.

PR #323, opened 2021-02-18
Only 2 groups for now

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T271904

### Problem Statement

We only want the control and trending groups for now.

### Solution

Removed the topics group.

Renamed the storage key so this group assignment is specific to this experiment.


medied (approved):
Will update #322 after this is merged

PR #332, opened 2021-02-26
Deploy the app and simulator for all branches

stephanebisson:
Phabricator Link: -

### Problem Statement

Wouldn't it be nice to be able to test any branch in the simulator without downloading the code?

### Solution

Deploy every branch to gh-pages

Example: https://wikimedia.github.io/wikipedia-kaios/sim-all-branches/sim.html

### Note

There is no automated way to clean up the old branch directories after they are deleted, I'll clean them up manually once in a while.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ jobs:
       - run:
           name: Deploy sim
           command: |
-            gh-pages --dist ./gh-pages --dotfiles -m "[ci skip] Update simulator"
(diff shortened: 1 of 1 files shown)


jpita:
Maybe we can use github actions to run a script or call an API when a branch is deleted?
Or at least to send us an email to clean the folder

stephanebisson:
> Maybe we can use github actions to run a script or call an API when a branch is deleted?
> Or at least to send us an email to clean the folder
> https://github.community/t/triggering-actions-on-branch-deletion/17743

Yes, we can look into that eventually

PR #333, opened 2021-03-01
Temporarily disable logging experiment group in InukaPageView

stephanebisson:
Phabricator -

### Problem Statement

We want to release a pre-experiment version of the app.

### Solution

Don't include the `tests` key in the InukaPageView experiment just yet.


hueitan (approved):
Yes

PR #336, opened 2021-03-04
Don't show Feed component when it is not available

hueitan:
Phabricator Link: -

### Problem Statement

It shows empty feed component 

### Solution

hide it

### Note

tested on https://wikimedia.github.io/wikipedia-kaios/T275621-experiment-config/sim.html to see empty feed component

### DIFF
--- a/src/components/Feed.js
+++ b/src/components/Feed.js
@@ -21,10 +21,10 @@ export const Feed = ({ lang, isExpanded, setIsExpanded, lastIndex, setNavigation
     }
   }, [trendingArticles])
 
-  return (
+  return trendingArticles.length ? (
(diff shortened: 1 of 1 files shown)


medied (commented):
I think this would conflict with #337. We have a `showFeed` boolean in Search, is that a better place to decide whether to show the feed or not

hueitan:
> I think this would conflict with #337. 

Yes, it has conflict with the loading state.

> We have a `showFeed` boolean in Search, is that a better place to decide whether to show the feed or not?

We will need to load api in the Feed component, don't see an easy way to bind the api response to `showFeed`, the easiest way to return `null` in this component

Closing this as this can be fixed in loading pr or another pr after the loading pr merged.

PR #341, opened 2021-03-04
Cypress intercept

jpita:

### Problem Statement

We were having issues with the user interface being slow to respond to cypress after a search.

### Solution

Using cy.intercept seems to be the best way to avoid this since we stub the call and have a local file with the response to the search API call.


jpita:
moving this to draft, needs more testing to figure out what's making the tests fail

hueitan (approved):
if this get faster, why not?

PR #343, opened 2021-03-09
Add cypress dashboard

jpita:
Add access to cypress dashboard

https://dashboard.cypress.io/projects/7izdto/runs

Super handy to trackdown flaky tests and to get an overview of our tests and metrics.

Also added a badge to the README that shows the test run status and numbers.

Also added test retries that hopefully makes our builds more reliable in case we have a flaky test.


hueitan (approved):
Not bad, nice to have 👍

PR #337, opened 2021-03-04
Add feed loading UI

medied:
https://phabricator.wikimedia.org/T271677

### Objective

Add feed loading UI

PR #346, opened 2021-03-15
Update trending articles page url

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T277041

### Problem Statement

The strategy for storing trending articles list has changed. It is now a single page per country.

### Solution

Update the url to remove the date and add the country code.

Also minor cleanup.


src/components/Feed.js
```diff
@@ -3,16 +3,18 @@ import { useEffect, useState } from 'preact/hooks'
...
   const i18n = useI18n()
+  const userCountry = getUserCountry()
 
   useEffect(() => {
-    const [request, abort] = getTrendingArticles(lang)
+    const [request, abort] = getTrendingArticles(lang, userCountry)
     request.then(setTrendingArticles)
     return abort
-  })
+  }, [lang, userCountry])
```
hueitan, line 17: Just wondering do we need these 2 state here? We will never reach here if one of them is null

hueitan (approved):
Nice!

src/components/Feed.js
medied: That's correct, realized that when working on loading state PR (https://github.com/wikimedia/wikipedia-kaios/pull/337/files#diff-ed6756f61e75bd6003f7ec1a246fa0444f9d4074a95da69e536e5ee39b6a09a4R20)

PR #350, opened 2021-03-16
Harden experiment config

stephanebisson:
Phabricator Link: -

### Problem Statement

We've seen the fetching of the experiment config fail from time to time in CI with "page.revisions is undefined". Wondering whaat's up with that.

### Solution

Log a warning with the content of the API response when this happens.

### Note

I hope this is temporary.


hueitan (approved):
OK

PR #353, opened 2021-03-17
Flag simulator as BETA

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T277257

### Problem Statement

The simulator is getting more popular but it may not work with all browsers config.

### Solution

Fix what we know about (sendBeacon) and add a disclaimer and some help.

### Note

Sim Link: https://wikimedia.github.io/wikipedia-kaios/sim-beta/sim.html

### DIFF
--- a/sim.html
+++ b/sim.html
@@ -11,10 +11,19 @@
 	<body>
 		<div class="container">
 		<div class="panel help">
-			<h2>BETA</h2>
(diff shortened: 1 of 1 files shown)


hueitan (approved):
`data-allowlist` is very cool!

PR #359, opened 2021-03-22
T277497: Add basic error handling for trending articles feed

medied:
https://phabricator.wikimedia.org/T277497

### Objective 

How should the app react when the api returns falsy/empty data. A potential solution is an error state to render in the feed UI, instead of just an empty feed (which is the current behavior).

We also need to we need to log it in our client-side errors stream (`sendErrorLog`)

### Note

Sim Link: https://wikimedia.github.io/wikipedia-kaios/T277497-feed-error/sim.html

### DIFF
--- a/src/api/getTrendingArticles.js
+++ b/src/api/getTrendingArticles.js
@@ -13,7 +13,7 @@ export const getTrendingArticles = (lang, country) => {
   return cachedFetch(url, data => {
     const page = data.query.pages[0]
     if (page.missing) {
-      sendErrorLog({ message: page.title, url })
+      sendErrorLog({ message: `There was an issue fetching '${page.title}', the full API response is: ${JSON.stringify(data)}`, url })
       return []
     }
     return JSON.parse(page.revisions[0].slots.main.content)
--- a/src/api/getTrendingArticles.js
+++ b/src/api/getTrendingArticles.js
@@ -1,4 +1,4 @@
-import { buildMwOrgApiUrl, cachedFetch } from 'utils'
+import { buildMwOrgApiUrl, cachedFetch, sendErrorLog } from 'utils'
 
 export const getTrendingArticles = (lang, country) => {
   const title = `Wikipedia_for_KaiOS/engagement1/trending/${lang}/${country.toLowerCase()}`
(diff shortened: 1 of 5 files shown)


src/api/getTrendingArticles.js
```diff
@@ -13,6 +13,7 @@ export const getTrendingArticles = (lang, country) => {
   return cachedFetch(url, data => {
     const page = data.query.pages[0]
     if (page.missing) {
+      sendErrorLog({ message: page, url })
```
hueitan, line 16: the page is an object, the message here should be `string`

hueitan: ah, that's fine as it will stringify later

hueitan:
it has `400` error when sending the error event, can you double check it?

src/api/getTrendingArticles.js
medied: Yea, does it make sense to pass in the whole `page` object though?

medied: Actually answering my own question, the `400` error seems to be related to `message` not being a string

medied:
Looks like `missing`, `ns`, and `title` are the available keys when there's an error. I'm gonna pass in `title` which is the actual string for now, but I wonder if we should have a more useful error message available

hueitan (approved):

src/api/getTrendingArticles.js
```diff
@@ -13,6 +13,7 @@ export const getTrendingArticles = (lang, country) => {
   return cachedFetch(url, data => {
     const page = data.query.pages[0]
     if (page.missing) {
+      sendErrorLog({ message: page.title, url })
```
stephanebisson, line 16: The `message` here needs to be the actual error message we will see in logstash. You can make it something like `Unable to fetch ${page.title} because blah...`. You can even happen `JSON.stringify(data)` at the end to make sure we have the full API response at our disposal to troubeshoot the issue.

stephanebisson (changes requested):

src/api/getTrendingArticles.js
medied: That's a perfectly great idea right there

PR #360, opened 2021-03-23
T278133: Display only the first 5 items in feed list

medied:
https://phabricator.wikimedia.org/T278133

## Objective

The list of trending articles fetched from mediawiki may contain more than 5 items but the app should only use the first 5.

### Note

Sim Link: https://wikimedia.github.io/wikipedia-kaios/T278133-feed-max/sim.html


stephanebisson (changes requested):
Slicing in the view means we are keeping the full list in memory and slicing it (copying it) every time the view renders. Doing it in the API layer would be cheaper.

PR #361, opened 2021-03-24
T271677: Remove setNavigation conflict

medied:
https://phabricator.wikimedia.org/T271677

### Problem Statement

Huei found this bug: if user selects an article from trending feed and then hits `backspace` to return to the feed, the `lastIndex` item is not rendered as selected (and instead the first item is highlighted 🐛). This is because a `setNavigation(1)` from a second `useEffect` is overriding the navigation logic for when the user returns from article. This bug was introduced in #337. 

### Solution

We can actually simplify the `useEffect` calls into one because they actually perform similar/complementary logic with `setNavigation`. The dependencies observed remain the same (`trendingArticles`, `loading`, `isExpanded`)

### Note

Sim Link: https://wikimedia.github.io/wikipedia-kaios/T271677-feed-loading-followup-lastIndex-bug/sim.html

### DIFF
--- a/src/components/Feed.js
+++ b/src/components/Feed.js
@@ -25,14 +25,10 @@ export const Feed = ({ lang, isExpanded, setIsExpanded, lastIndex, setNavigation
     if (lastIndex) {
       setIsExpanded(true)
       setNavigation(lastIndex)
-    }
-  }, [trendingArticles])
-
-  useEffect(() => {
-    if (isExpanded) {
+    } else if (isExpanded) {
       setNavigation(1)
     }
-  }, [loading, isExpanded])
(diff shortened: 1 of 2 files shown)


hueitan (changes requested):
it works when go to the article and back,
but going to tips/settings and back doesn't work yet

hueitan (commented):
another option would be setting the last feed index every time when it goes up and down

listen to the list index change instead of adding the handler in every possible event.

```js
useEffect( () => {
    handleLastFeedIndex();
}, [current] )
```

but the current solution also fine

stephanebisson (approved):

PR #364, opened 2021-04-17
Search only for confirmed characters

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T246786

### Problem Statement

On most KaiOS devices, keys are grouped 4 per buttons (2abc, 3def, ...). To type most of the characters the user has to cycle through other characters. Triggering searches for all the intermediate characters is wasteful. 

### Solution

Trigger search for input events that are not part of a composition session.

Update the simulator to behave like the device (as far as I can tell).

### Note

THIS NEEDS TO BE TESTED AS MUCH AS POSSIBLE ON ALL AVAILABLE DEVICES.

Sim Link: https://wikimedia.github.io/wikipedia-kaios/T246786/sim.html


hueitan (dismissed):
Works well in Nokia Flip! 💯

src/components/Search.js
```diff
@@ -97,8 +98,9 @@ export const Search = () => {
     }
   }
 
-  const onInput = ({ target }) => {
-    if (isOnline) {
+  const onInput = ({ target, isComposing }) => {
```
medied, line 101: Nice, this is a TIL for me: https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/isComposing

medied:
I noticed a minor inconsistency from what's in `main`: if you type something into search bar, then select any of the results from the list, then press backspace key to return from the article to the search results list, you end up with the same list but the search bar is empty (with the "Search Wikipedia" placeholder)

stephanebisson:
Good catch Eduardo. Should be addressed in the latest commit. Please test carefully again.

hueitan (approved):
it works for me

medied (approved):

jpita:
when I chose the predictive text option on the nokia phone, it only searches when I chose the text.
we can't control this, right? this is when the system sends the input?

jpita:
on jio2, when I search for "albuquerque", this is how many calls we do:
image
should this task include a delay for jio2 as well or just for jio1/banana phone type of keyboard?

stephanebisson:
> when I chose the predictive text option on the nokia phone, it only searches when I chose the text.
> we can't control this, right? this is when the system sends the input?

I need to test this. In general it should behave similarly to the single character selection: once a suggested word is selected, it become officially part of the query and a search should be triggered. While the user is toggling between suggested words with the arrow keys, no search is triggered because the work is not confirmed.

We receive all inputs and we know if it's part of a composing session or if it's a confirmed input so we have a decent amount of control.

jpita:
> I need to test this. In general it should behave similarly to the single character selection: once a suggested word is selected, it become officially part of the query and a search should be triggered. While the user is toggling between suggested words with the arrow keys, no search is triggered because the work is not confirmed.
> 
> We receive all inputs and we know if it's part of a composing session or if it's a confirmed input so we have a decent amount of control.

what u mentioned is the current behaviour, so I guess it's ok.
the search is only started when I select the word

stephanebisson:
> on jio2, when I search for "albuquerque", this is how many calls we do [...] should this task include a delay for jio2 as well or just for jio1/banana phone type of keyboard?

All those characters are confirmed so we want to search for all of them. In the context of this task this is the right behavior.

PR #365, opened 2021-04-27
T277497: Guard for undefined page

medied:
### Objective

QA brought up if for whatever reason the response of `getTrendingArticles` comes back with `data.query.pages[0]` as `undefined`, a `TypeError` is thrown. This adds a guard against that. 

See https://phabricator.wikimedia.org/T277497#7038271 for more

### DIFF
--- a/src/api/getTrendingArticles.js
+++ b/src/api/getTrendingArticles.js
@@ -12,7 +12,7 @@ export const getTrendingArticles = (lang, country) => {
   const url = buildMwOrgApiUrl(params)
   return cachedFetch(url, data => {
     const page = data.query.pages[0]
-    if (page.missing) {
+    if (page && page.missing) {
       sendErrorLog({ message: `There was an issue fetching '${page.title}', the full API response is: ${JSON.stringify(data)}`, url })
       return []
     }


src/api/getTrendingArticles.js
```diff
@@ -12,7 +12,7 @@ export const getTrendingArticles = (lang, country) => {
   const url = buildMwOrgApiUrl(params)
   return cachedFetch(url, data => {
     const page = data.query.pages[0]
-    if (page.missing) {
+    if (page && page.missing) {
```
hueitan, line 15: I don't see a good reason to have the extra guard here, what if `data.query` is undefined? 

a better solution would be having try catch form and failureCallback in the promise, but we don't apply this to all the api layer.

if the api changes the response object, we will need to update it too.

jpita: so I changed L14 to
```const page = data.query.pages[-1]```
and I get this error
image

I feel like this is a rabbit hole though, since this won't probably happen in prod?
Shall we just ignore this?

medied: Fine by me

PR #369, opened 2021-06-09
Script to capture memory usage, and other stats, in real time

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T283725

### Problem Statement

It was reported that the KaiOS app was using too much memory. The memory tab in Firefox dev tools can take memory snapshots at any given time but it cannot show live usage.

### Solution

This script can be left running while the app is being used to generate a log of memory usage and other metrics.

### Note

Sim Link: N/A

### DIFF
--- a/scripts/stats.sh
+++ b/scripts/stats.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+BASEDIR=$(dirname "$0")
+$BASEDIR/stats_header.sh
+$BASEDIR/stats_line.sh
--- a/scripts/stats_header.sh
+++ b/scripts/stats_header.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+# USS and PSS are the important columns for memory usage analysis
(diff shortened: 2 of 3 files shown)


medied:
Should this work on a banana (non-Jio) phone? I plug in my device, shows up with `adb devices`, and see this when running the script and using the app in the phone:

```
Eduardos-MacBook-Pro:scripts medied$ ./stats.sh
TIME NAME PID PPID CPU(s) NICE USS PSS RSS SWAP VSIZE ADJ SCORE_ADJ USER
17:39:30
17:39:30
17:39:30
…
17:40:28
17:40:28
17:40:28
./stats_line.sh: fork: Resource temporarily unavailable
```

What am I missing?

jpita:
I see this
image

what values should I be looking out for?

stephanebisson:
> Should this work on a banana (non-Jio) phone? 

No

stephanebisson:
Whether we find a memory issue or not I will merge this so we keep those scripts around.

PR #370, opened 2021-06-11
Convert lang code nb to no for Norwegian (Bokmal)

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T280274

### Problem Statement

The langlinks API returns the language code 'nb' for Norwegian (Bokmal) but Wikipedia uses 'no' as the subdomain. 

### Solution

Convert the API response from 'nb' to 'no'

To test, I used `en/Earth`, then use the language button to switch to `no/Jorden`

### Note

Sim Link: https://wikimedia.github.io/wikipedia-kaios/no-nb/sim.html

### DIFF
--- a/src/api/getLanglinks.js
+++ b/src/api/getLanglinks.js
@@ -1,8 +1,13 @@
 import {
   buildMwApiUrl, cachedFetch,
   isSupportedForReading, getDirection,
-  prioritizedList
+  isPrioritized
 } from 'utils'
+
+const convertToWikiLang = lang => {
+  // The lang link endpoint return 'nb' but the wiki subdomain is 'no'
+  return lang === 'nb' ? 'no' : lang
+}
 
 export const getLanglinks = (lang, title) => {
   const params = {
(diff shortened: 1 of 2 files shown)


src/api/getLanglinks.js
```diff
@@ -1,9 +1,14 @@
 import {
   buildMwApiUrl, cachedFetch,
   isSupportedForReading, getDirection,
-  prioritizedList
+  isPrioritized
 } from 'utils'
 
+const convertToWikiLang = lang => {
+  // The lang link endpoint return 'nb' but the wiki subdomain is 'no'
```
Nikerabbit, line 8: This is not future proof as there is likely going to be more exceptions in the future.

Why are you not using the `url` prop, e.g. https://en.wikipedia.org/wiki/Special:ApiSandbox#action=query&format=json&prop=langlinks&titles=London&llprop=url%7Clangname%7Cautonym&lllimit=500

src/api/getLanglinks.js
```diff
@@ -16,34 +21,37 @@ export const getLanglinks = (lang, title) => {
...
-            dir: getDirection(item.lang)
-          }
-        })
-    )
+      .filter(item => isSupportedForReading(item.lang))
+
+    // Sort languages prioritized first, and then
+    // in alphabetical order within the prioritized
+    // and non-prioritized groups.
+    allLanguages.sort((a, b) => {
```
hueitan, line 39: in app language setting, we have the different sort algorithm, should we do the same there?

src/api/getLanglinks.js
```diff
@@ -16,34 +21,37 @@ export const getLanglinks = (lang, title) => {
...
-          return {
-            title: item.autonym,
-            langname: item.langname,
-            lang: item.lang,
-            description: item.title,
-            dir: getDirection(item.lang)
-          }
-        })
-    )
+      .filter(item => isSupportedForReading(item.lang))
```
hueitan, line 34: `filter` should go first then `map`, so unsupported language don't need to go through `map`

src/api/getLanglinks.js
stephanebisson: Good idea. I did it in the latest commit.

src/api/getLanglinks.js
stephanebisson: I prefer to map first so we filter with the converted land code.

src/api/getLanglinks.js
stephanebisson: The main difference is over there, both `prioritizedLanguages` and `allLanguages` are used so it makes sense to build them separately.

hueitan (approved):

medied (approved):

PR #374, opened 2021-08-09
Open links with the external browser

stephanebisson:
Phabricator Link: N/A

### Problem Statement

The app needs to open external pages in some cases like the privacy policy and media pages. We do so with `window.open()`. It works well but the memory used by that window is reported as part of the application.

### Solution

Open external links with a WebActivity so that the memory is reported as the browser application.

### Note

Sim Link: https://wikimedia.github.io/wikipedia-kaios/open-external/sim.html


medied:
Working well from my end both on desktop sim and actual device. The only differences in behavior I noticed are:

* when you open a new window and then press the back key, 
  * with `openExternal` it asks you to quit the browser and lands you back on the phone app menu, 
  * with `open.window` it lowers/closes the browser wrapper and lands you where you left off from Wikipedia app. 
* also, something that could be argued as positive: 
  * with `openExternal` a user can now land on an external page and then click on another link and keep on browsing normally, and pressing the back key would take you back to the previous browser page as expected. 
  * with `open.window` pressing the back key always takes you back to Wikipedia app without warning

These are definitely not a blocker for me, specially if this helps with memory consumption.

hueitan (approved):
it makes sense.

stephanebisson:
@medied you're right, the experience is different because with the WebActivity we shell out to a different app and the interaction between apps is not perfect, especially with regards to the back button. 

For the sake of a significant improvement in perceived memory usage, I'll go ahead and merge this.

PR #372, opened 2021-07-16
T286725: Use screen-sized images

medied:
https://phabricator.wikimedia.org/T286725

To reduce memory usage, we can grab just the title strings with `media-list` and then use those to fetch the screen-sized image directly from `Gallery`. Tested on my device: a bit slower (yours too?) but lighter on memory. Is it worth adding a little loading UI in between images?

PR #375, opened 2021-08-09
Fix Gallery/About

stephanebisson:
Phabricator Link: https://phabricator.wikimedia.org/T286725

### Problem Statement

Follow-up to #372 
The gallery/about feature doesn't have access to the media info and stays in loading state.

### Solution

Add `mediaInfo` as a softkey dependency so that the handlers are updated when the media info becomes available.

### Note

Sim Link: https://wikimedia.github.io/wikipedia-kaios/mediaInfo/sim.html


medied (approved):
Testing looks good. Good catch! 👏

PR #376, opened 2021-08-12
test for mediaInfo issue

jpita:
Phabricator Link: https://phabricator.wikimedia.org/T288493

following #375 , I created a test to make sure this doesn't come back again


cypress/integration/article-view-spec.js
```diff
@@ -56,12 +56,9 @@ describe('Article view', () => {
     goToCatArticle()
     articlePage.selectOptionFromActionsMenu('gallery')
     articlePage.galleryImage().should('be.visible')
-    articlePage.galleryImage().invoke('attr', 'src').then((src) => {
-      cy.rightArrow()
-      articlePage.galleryImage().should('not.have.attr', 'src', src)
-    })
     cy.enter()
     articlePage.galleryPopupHeader().should('be.visible')
+    articlePage.galleryPopupDescriptionText().should('be.visible').should('have.text', 'A montage of cat pictures using images from Wikimedia creators')
```
jpita, line 61: @medied can u point me to the call/url/api we use to get this information about the image?
this test passes now but if the image or description changes on wikipedia then it will fail, I'd like to avoid that by intercepting the call to get the real data that is going to be shown

jpita: cc @hueitan and @stephanebisson

hueitan: take this as the example
```
https://commons.wikimedia.org/w/api.php?format=json&formatversion=2&origin=*&action=query&prop=imageinfo&iiextmetadatafilter=License%7CLicenseShortName%7CImageDescription%7CArtist&iiextmetadatalanguage=en&iiextmetadatamultilang=1&iiurlwidth=240&iiurlheight=320&iiprop=url%7Cextmetadata&titles=File%3AWild-domestic-hybrid_cat_skulls.png
```
image

PR #379, opened 2021-09-10
Various optimization

stephanebisson:
This PR was supposed to be about optimizing the article inline images but I couldn't do anything that makes a difference. In the process, I made other performance optimization that help with speed that are worth reviewing and maybe merging.

- Use a single document object per section to prepare the api output
- Get rid of unused article preview in getArticle()
- Add memo() to a lot of components to make them pure (they won't rerender if the props haven't changed)

Sim Link: https://wikimedia.github.io/wikipedia-kaios/T290579/sim.html


hueitan (approved):
I don't see any problem of the changes. 👍

PR #380, opened 2022-06-03
Add the EndCall event to confirm the exit app

hueitan:
Phabricator Link: -

### Problem Statement

End Call will exit the app without confirm dialog

### Solution

Add the Confirm Dialog Globally when user land on Search page

Tested on

 - [x] Jio device (device has back and endcall in the seperate key)
 - [x] KaiOS device, flip phone (device has back/end call in the seperate key)
 - [x] KaiOS device (device has back/end call in the same key)

### Note

Sim Link: - (only available on device)


src/components/Search.js
```diff
@@ -153,6 +153,22 @@ export const Search = () => {
...
+      if (key === 'EndCall') {
+        onExitConfirmDialog()
+        e.stopPropagation()
+        e.preventDefault()
+      }
+    }
+
+    document.addEventListener('keydown', onKeyEndCall)
+  }, [])
+
```
hueitan, line 171: we can also apply this only to Jio if KaiOS doesn't need it.

src/hooks/useGlobalEndCallKey.js
```diff
@@ -0,0 +1,17 @@
...
+      const key = e.key.toString()
+
+      if (key === 'EndCall') {
+        callback()
+        e.stopPropagation()
+        e.preventDefault()
+      }
+    }
+
+    document.addEventListener('keydown', onKeyEndCall)
```
stephanebisson, line 15: This function needs to return a unregister function

stephanebisson (commented):
It works well but I'm curious if it could be integrated to the useSoftkeys already in place in this component.

src/hooks/useGlobalEndCallKey.js
hueitan: I'm wonder if the unregister function is needed, because this is the "global" event that not going to remove in any places, 

and I'm thinking is that better to have a guard to make sure this define only once.

hueitan:
> It works well but I'm curious if it could be integrated to the useSoftkeys already in place in this component.

I was thinking the same, the `useSoftkeys` is defined based on each origin, this global event belongs to all, we can easilly having `onEndCallKey` in all the origin/pages but it's way too much to repeat it.

Perhaps having a "global" optioin in useSoftkey to make it a global event?

stephanebisson:
Ah, I see what you mean by global. That's also why you don't have a unregister function in the hook. But it looks like it may re-register the handler every time you land on the search screen. Something truly global should be in App

hueitan:
> Ah, I see what you mean by global. That's also why you don't have a unregister function in the hook. But it looks like it may re-register the handler every time you land on the search screen. Something truly global should be in App

1. We can't use `App` as it throws the `Context` error, It could be the reason popupContext provider not been defined yet, therefore moving the event to `Router`.

2. I also didn't touch the useSoftkey and SoftkeyReducer, we need to add the this global into the stack every time when dispatch, this will bring the reducer even more complex, i would suggest wait until we see more global cases.

PR #382, opened 2022-06-23
Disable the Experimental Group (not removing the entire code)

hueitan:
Phabricator Link: https://phabricator.wikimedia.org/T311142

### Problem Statement

Disable the Feature

### Solution

return false but not removing the entire code yet


stephanebisson (changes requested):
The goal is to prevent the app from fetching the experiment config from mw.org. It looks like with the change it will still do it but ignore the details of the config and always return false.

stephanebisson (approved):

